PR #326, opened 2026-03-15
New Permission System

### LINKED ISSUE #303: Req 1: Each user can have one or many permissions.
Add new database table
```
permissions:
user_id | permission
```

- user_id should link to `localuser`
- permission is enum with values LECTURER, MAINTAINER, ADMIN

-----
See #301

### LINKED ISSUE #304: Req 3: Users with the "maintainer" permission can manage prompts

### LINKED ISSUE #305: Req 4: Users with the "maintainer" permission can manage users
(edit permissions, block/delete user, etc.)

### LINKED ISSUE #306: Req 5: Users with the "admin" permission implicitly have all other permissions

SamuABC:
resolves #303, resolves #304, resolves #305, resolves #306

# Changes
- added a Permission table that stores an enum value (LECTURER, MAINTAINER or ADMIN) that is connected to a localuser_id. On migration, all current local ADMINs get the global ADMIN permission as well
- renamed the lecture ADMIN role to OWNER (I didn't rename thinks like "admin settings page" for now because I think we will move a lot of things around anyway later)
- implemented new protection headers that check if the user has the needed local role **OR** one of the allowed global roles. These new protection headers should later also check if the user is part of the lecture which the page belongs to.
- global ADMINs get access to everything (the protection header automatically adds the global ADMIN to the list of allowed permissions. Nevertheless, I included the ADMIN role in the call of the protection headers that use the `allowed_permissions` argument. This makes it clear for the reader of the code that the ADMIN is also allowed)
- MAINTAINERS have access to the prompt management and to the user management
- the admin_link navigation bar now also shows for MAINTAINERS and global ADMINs


aitutor/auth/protection.py
```diff
@@ -63,17 +97,40 @@ def protected_page():
     return decorator
 
 
-def state_require_role_at_least(required_role):
+def state_require_role_or_permission(
```
luator: same as above

aitutor/auth/state.py
```diff
@@ -106,3 +106,23 @@ def user_role(self) -> UserRole | None:
...
+        ):
+            return []
+        with rx.session() as session:
+            permissions = session.exec(
+                select(Permission.permission).where(
+                    Permission.user_id == self.authenticated_user.id
+                )
+            ).all()
+            # Cast each permission string to GlobalRole enum
+            return [GlobalRole(perm) for perm in permissions]
```
luator: Not sure but is this cast actually needed?  I would have expected that SQLAlchemy takes care of that.

luator (changes requested):
Sorry for the late review, the last few days have been very busy.

I realised that we didn't plan anything yet on where/how to edit those user permissions.  I'm going to add an issue for this to the milestone.  It could be added in this PR or a separate one, as you prefer.

aitutor/models.py
```diff
@@ -37,7 +37,17 @@ class UserRole(IntEnum):
 
     STUDENT = 1
     TUTOR = 2
-    ADMIN = 3
+    OWNER = 3
```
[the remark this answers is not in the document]
luator: Result from discussion: The current migration script should be tested with postgres, to see if the renaming works there.

luator (approved):
Thanks a lot!

PR #333, opened 2026-04-04
Andrew/token analyzer

### LINKED ISSUE #167: Track and analyse tokens used by OpenAI API requests
This is just a high-level issue.  Actual work packages are in sub-issues.

General idea: When making requests to the OpenAI API, keep track of the tokens used, so we can see and intervene if there is some excessive usage.

### LINKED ISSUE #170: Add page to analyse token usage
Add a page (access level "teacher") that shows an analysis of used API tokens.

It should contain (maybe organised in tabs or the like?)

- the overall number of used tokens across all exercises and users
- for a selected exercise, the total number of used tokens and maybe a sorted bar plot with tokens used per user.
- for a selected user, the total number of used tokens and tokens used per exercise (as plot or table?)

Andrewtuebingen:
# Pull Request: Token Analyzer Page

## Overview

This PR introduces a new **Token Analyzer page** for admins to monitor token usage.

- Adds two analysis views:
  - User-based token usage
  - Exercise-based token usage
- Includes filtering (user / exercise) with prefix search and "All" option
- Displays data in both tables and bar charts
- Uses aggregated queries to calculate total tokens
- Adds dynamic chart scaling (bar size, width, ticks) for large datasets

Overall, this enables better visibility into token consumption and helps identify heavy usage patterns.


aitutor/pages/token_analyzer/components.py
```diff
@@ -0,0 +1,367 @@
...
+) -> rx.Component:
+    """Reusable selector row with prefix-search input and clear button."""
+    return rx.hstack(
+        rx.box(
+            rx.text(
+                label_text + ":",
+                size="3",
+                weight="medium",
+            ),
+            width="80px",
```
luator: I think it's better to avoid pixel values for the sizes but rather use something like `em`, that is based on the text size and should thus scale with different monitor resolutions.
Also applies to the other elements below.

aitutor/pages/token_analyzer/components.py
```diff
@@ -0,0 +1,367 @@
...
+            width="100%",
+            height=320,
+            min_width=TokenAnalyzerState.chart_min_width,
+        ),
+        width="85vw",
+        overflow_x="auto",
+    )
+
+
+def token_analyzer_exercise_bar_chart() -> rx.Component:
```
luator, line 243: This and the function above seem to be mostly the same, just using different labels and data sources.  Would it make sense to have a helper function with the plot structure that takes these values as arguments?  Or is there some fundamental difference which I missed?

aitutor/pages/token_analyzer/components.py
```diff
@@ -0,0 +1,367 @@
...
+                    name=f"{TOOLTIP_PREFIX_SECONDARY}{LanguageState.token_usage}",
+                    fill=rx.color("indigo", 8),
+                    bar_size=TokenAnalyzerState.user_bar_size,
+                ),
+                rx.recharts.bar(
+                    data_key="username",
+                    name=f"{TOOLTIP_PREFIX_PRIMARY}{LanguageState.user}",
+                    fill="transparent",
+                    bar_size=0,
+                ),
```
luator, line 216: I'm trying to understand the purpose of this function.  Does it add an invisible bar at the same x-position so that the username shows up in the hover tooltip of the bar?
Might be good to add a brief comment to explain this, because just from looking at the code it can be a bit confusing.

aitutor/pages/token_analyzer/state.py
```diff
@@ -0,0 +1,387 @@
...
+    exercise_title: str
+    tokens_used: int
+
+
+class TokenAnalyzerState(SessionState):
+    """State for the token analyzer page."""
+
+    table_rows: list[TableRow]
+    chart_data: list[dict[str, int | float | str]]
+    chart_ticks: list[int]
```
luator: Are those for the user view?  If yes, please prefix with `user_` so it's consistent with the corresponding `exercise_...` fields below.

aitutor/pages/token_analyzer/state.py
```diff
@@ -0,0 +1,387 @@
...
+        self.exercise_bar_size = 3
+        self.user_options = []
+        self.selected_user_name = ALL_USERS_OPTION
+        self.user_filter_query = ""
+        self.user_bar_size = 3
+
+    @rx.var
+    def all_option_label(self) -> str:
+        """Localized label for the 'all' select option."""
+        return "Alle" if self.language == Language.DE else "All"
```
luator: Please use the `BackendTranslations` class in `language_state.py` for this (so all translations are in one place).

aitutor/pages/token_analyzer/state.py
```diff
@@ -0,0 +1,387 @@
...
+    ) -> list[str]:
+        """Filter option list by prefix (case-insensitive), keeping the all option."""
+        normalized_query = query.strip().lower()
+        if not normalized_query:
+            return options
+
+        filtered = [
+            option
+            for option in options
+            if option == all_option or option.lower().startswith(normalized_query)
```
luator: I wonder if it would be more useful to check if the search string appears anywhere in the string instead of just checking the start.  What do you think?

Code-wise it could be done like this:

```suggestion
            if option == all_option or normalized_query in option.lower()
```

aitutor/pages/token_analyzer/state.py
```diff
@@ -0,0 +1,387 @@
...
+    selected_user_name: str = ALL_USERS_OPTION
+    user_filter_query: str = ""
+    user_bar_size: int = 3
+
+    @rx.event
+    def set_selected_exercise_name(self, exercise_name: str):
+        """Set selected exercise filter and reload the token rows."""
+        self.selected_exercise_name = (
+            ALL_EXERCISES_OPTION
+            if exercise_name == self.all_option_label
```
luator, line 63: I think this kind of handling of the "all" function will somewhat break if there is an exercise called "All".  It would be nicer if we could get an id or index from the `rx.select` that is independent of the actual display value.  But at least from what I saw in the documentation this is not possible :(.  So I guess we have to keep it like that.

luator (changes requested):
Thanks, looks already pretty good.  I just have some comments on implementation details.

aitutor/pages/token_analyzer/components.py
Andrewtuebingen: Yes, exactly. It adds an invisible series so the tooltip can include the username/exercise title for the hovered rank. I’ve added a short comment to make that clearer in the code.

aitutor/pages/token_analyzer/components.py
Andrewtuebingen: no you are right, thanks

aitutor/pages/token_analyzer/state.py
Andrewtuebingen: should we somehow make a function that bans specific exercise names like "All"? But yea i dont think that someone will call the exercise this name.

aitutor/pages/token_analyzer/state.py
Andrewtuebingen: I like the idea, its more forgiving.

aitutor/pages/token_analyzer/state.py
luator: I don't think we need to care about it too much.  I also don't think that an exercise called "All" is likely to happen. Even if it would, the only issue would be that filtering for that specific exercise doesn't work, which I'd consider a rather minor problem.

luator (approved):
Thanks!

PR #336, opened 2026-04-21
fix: Update permission check decorators for token_analyzer

luator:
The token analyzer and the change that affected the permission decorators were merged in parallel, so this was not caught by checks or testing of the individual pull requests.

Luckily the pyright check run on main after the merge immediately revealed the issue.  I'll see if I can change the repo settings to avoid this in the future.


Andrewtuebingen (approved):
Great, thanks for the changes!

PR #337, opened 2026-04-21
Increase default token limit to 100k

luator:
It turned out that the default limit of 30k was a bit too low for actual use cases.  Increase it to 100k.

### DIFF
--- a/aitutor/models.py
+++ b/aitutor/models.py
@@ -229,7 +229,7 @@ class Config(SQLModel, table=True):
     course_name: str
     impressum_text: str
     registration_code: str
-    exercise_token_limit: int = Field(default=30000)
+    exercise_token_limit: int
 
     def __repr__(self):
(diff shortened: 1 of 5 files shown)


luator:
The failing checks are caused by an unrelated issue, which is fixed in #336.  So they should be resolved one that PR is merged.

Andrewtuebingen (changes requested):
Shouldnt we also change the other lines with 30k defaults?

Image

luator:
> Shouldnt we also change the other lines with 30k defaults?

Good catch!  Since the default values in the code are anyway not really used (they get overwritten right in the beginning by the `on_load` methods), I set them to zero instead, which is more generic.
From the database I removed the default completely, as to my understanding it is always set explicitly from the `default_config.toml` (or am I missing some cases where the default in the db is needed?)

Andrewtuebingen (approved):
Thanks!

PR #343, opened 2026-05-04
Andrew/editing permission

### LINKED ISSUE #309: Req 8: Users can see a list of all lectures

### LINKED ISSUE #310: Req 9: Users can join a lecture with a password
The password is set by the lecturer in the lecture properties.
When joining a lecture, the default role is "Student".

---
See #301

### LINKED ISSUE #312: Req 11: Users can see a separate list of all lectures they joined (to easily find them)
---
See #301

### LINKED ISSUE #313: Req 18: In-lecture role "Owner" can edit lecture properties (name, description, etc.)
---
See #301

### LINKED ISSUE #324: Req 7: Users can register without a registration code

### LINKED ISSUE #328: New Lecture table
New table "lecture" to save lectures:
- id
- check_conversation_prompt
- lecture_information_text
- lecture_name
- registration_code

And new table "link_user_lecture" with fields
- lecture_id
- user_id (referring to localuser.id)
- user role (enum with values STUDENT, TUTOR, OWNER)

Make `lecture_id` and `user_id` a compound primary key

### LINKED ISSUE #329: Add Page to Add New Lectures
- A page similar to the `Configuration` page but to add/edit (new) lectures
- The fields to be filled are the fields from the lecture table.
- Have 2 urls: `edit_lecture/42` or `edit_lecture/new`
   -  if existing lecture: load existing values + show different submit button name
   - if new lecture: show different submit button name

### LINKED ISSUE #339: delete lecture
in edit_lecture

Andrewtuebingen:
## What changed

This PR adds the main lecture management flow around three pages: **My Lectures**, **All Lectures**, and **Edit Lecture**. Together, these pages let users browse lectures, join available lectures, manage their own lectures, and create or update lecture details.

## My Lectures

- Added a dedicated page that shows the lectures visible to the current user.
- Users can search lectures by name and filter them by their role: owner, tutor, student, or all.
- Global admins can also see lectures they have not joined yet and edit them!
- Users with lecturer or admin permissions have a link button on this page to edit lecture to create new lectures
- Lecture owners and admins get an edit action directly in the table.

## All Lectures

- Added a page where students can browse all available lectures.
- Lectures are searchable by name and displayed together with lecturer information.
- Users can expand a lecture row to see additional lecture details.
- Students can join lectures directly from the table.
- If a lecture has a registration code, the join dialog asks for it before allowing the user to join.
- The join flow checks for missing lectures, duplicate joins, and invalid registration codes before creating the user-lecture link.
- Direct join links are supported by reading the lecture id from the route and opening the join dialog automatically for lecturer to share the link with students

## Edit Lecture

- Added a shared create/edit page for lectures.
- Lecturers can create new lectures (also Admin), while existing lectures can only be edited by owners or admins.
- The form supports lecture name, lecturer name, optional registration code, lecture information text, and the check-conversation prompt.
- New lectures automatically assign the creator as the lecture owner.
- The save logic validates required fields and prevents duplicate lecture names.
- Added unsaved-change handling, including a visual warning and disabled save/discard buttons when there is nothing to save.
- Added a button for copying a direct join link for existing lectures.
- Added password-confirmed lecture deletion for existing lectures, with proper permission checks and feedback messages.

## Access and permissions

- My Lectures and All Lectures are available to users with the student role.
- Edit Lecture is limited to users with lecturer permission and admins.
- Editing and deleting existing lectures is restricted to lecture owners or global admins.
- Admins get broader visibility in My Lectures (because he is the explicit owner) while regular users only see their joined lectures.


aitutor/pages/all_lectures/state.py
```diff
@@ -0,0 +1,272 @@
...
+    @rx.event
+    @state_require_role_or_permission(required_role=UserRole.STUDENT)
+    def on_load(self):
+        """Initialize the page state."""
+        self.global_load()
+        self._reset_page_state()
+        self.load_lectures()
+
+        lecture_id_param = self._get_route_lecture_id_param()
+        self.lecture_id_param = lecture_id_param
```
luator: Shouldn't this line rather be after the `if` below?

aitutor/pages/all_lectures/state.py
```diff
@@ -0,0 +1,272 @@
...
+
+        self.lectures = self._serialize_lectures(lectures)
+
+    def _get_route_lecture_id_param(self) -> str:
+        """Return the lecture id route segment or an empty string."""
+        path_segments = self.router.url.path.rstrip("/").split("/")
+        last_segment = path_segments[-1]
+        if last_segment == routes.ALL_LECTURES.split("/")[-1]:
+            return ""
+        return str(last_segment)
```
luator: Did you already take a look at Reflex's [dynamic routing](https://reflex.dev/docs/pages/dynamic-routing/) as potential alternative to manually parsing the url?  It would be more convenient to use but at least on first glance, it doesn't seem to support optional parameters :/

aitutor/pages/edit_lecture/page.py
```diff
@@ -0,0 +1,23 @@
...
+from aitutor.auth.protection import page_require_role_or_permission
+from aitutor.models import GlobalPermission
+from aitutor.pages.edit_lecture.components import edit_lecture_form
+from aitutor.pages.navbar import with_navbar
+from aitutor.pages.navbar_lectures import with_lectures_navbar
+
+
+@with_navbar(routes.LECTURES)
+@with_lectures_navbar(routes.EDIT_LECTURE + "/new")
+@page_require_role_or_permission(allowed_permissions=[GlobalPermission.LECTURER])
```
luator, line 15: Maybe good enough for this PR, so just meant as a reminder: The "lecturer" permission is needed to create *new* lectures, however, anyone could be assigned as "owner" of a an existing lecture.  Thus for editing an existing lecture, a different permission check will be needed.

aitutor/pages/edit_lecture/state.py
```diff
@@ -0,0 +1,290 @@
...
+        self.lecture_id_param = lecture_id_param
+
+        if lecture_id_param == "new":
+            self._reset_form(
+                check_conversation_prompt=DEFAULT_CHECK_CONVERSATION_PROMPT,
+            )
+            return
+
+        try:
+            lecture_id = int(str(lecture_id_param))
```
luator: Is the `str` really needed here?

aitutor/pages/edit_lecture/state.py
```diff
@@ -0,0 +1,290 @@
...
+        lecture_name = self._normalized_lecture_name()
+        if not lecture_name:
+            return rx.window_alert(BT.enter_lecture_name(self.language))
+
+        lecturer_name = self._normalized_lecturer_name()
+        if not lecturer_name:
+            return rx.window_alert(BT.enter_lecturer_name(self.language))
+
+        if self.authenticated_user is None or self.authenticated_user.id is None:
+            return rx.redirect(routes.LOGIN)
```
luator: This check should better be at the top of the method.

aitutor/pages/edit_lecture/state.py
```diff
@@ -0,0 +1,290 @@
...
+        self.delete_confirmation_password = ""
+
+    def _apply_lecture_to_state(self, lecture: Lecture) -> None:
+        """Copy persisted lecture values into the page state."""
+        self.current_lecture_id = lecture.id
+        self.lecture_name = lecture.lecture_name
+        self.lecturer_name = lecture.lecturer_name
+        self.registration_code = lecture.registration_code
+        self.lecture_information_text = lecture.lecture_information_text
+        self.check_conversation_prompt = lecture.check_conversation_prompt
```
luator, line 242: Would it be possible to directly store the `Lecture` object in the state (i.e. something like `self.lecture = lecture`)?  It seems it would simplify the code but I might miss some issues it would cause (or maybe it simply doesn't work like that in Reflex states?).

aitutor/pages/edit_lecture/state.py
```diff
@@ -0,0 +1,290 @@
...
+        lecture.registration_code = self.registration_code
+        lecture.lecture_information_text = self.lecture_information_text
+        lecture.check_conversation_prompt = self.check_conversation_prompt
+
+    def _get_route_lecture_id_param(self) -> str:
+        """Return the lecture id route segment or 'new' for the create page."""
+        lecture_id_param = self.router.url.path.rstrip("/").split("/")[-1]
+        if lecture_id_param == routes.EDIT_LECTURE.split("/")[-1]:
+            return "new"
+        return str(lecture_id_param)
```
luator: If I didn't miss anything, the edit_lecture page always has an id parameter (either "new" or an actual id), is that right?  If yes, here dynamic routing should work and would remove the need for manually parsing (see my comment above on the all_lectures page).

aitutor/pages/manage_users/state.py
```diff
@@ -185,6 +188,23 @@ def delete_user(self, user_id: int):
...
+
+                lecture_ids_to_delete: list[int] = []
+                for owner_link in owner_links:
+                    if owner_link.lecture_id is not None:
+                        lecture_ids_to_delete.append(owner_link.lecture_id)
+
+                for lecture_id in lecture_ids_to_delete:
+                    lecture = session.get(Lecture, lecture_id)
+                    if lecture is not None:
+                        session.delete(lecture)
```
luator: Silently deleting all lectures the user is owner off seems quite dangerous to me.  I'd rather block deleting the user in this case with some useful error message.  So if an admin want's to delete the user, they first have to manually handle the lectures (either delete them or assign another owner).

It might also be possible that all the lectures already have a second owner, in which case the user could safely be deleted.

In the future, we could add something to list all the lectures the user is owner of, but I think that is not important for now.

aitutor/pages/my_lectures/components.py
```diff
@@ -0,0 +1,210 @@
...
+import aitutor.routes as routes
+from aitutor.language_state import LanguageState as LS
+from aitutor.models import Lecture, LectureRole
+from aitutor.pages.my_lectures.state import LectureWithRole, MyLecturesState
+
+
+def lecture_role_text(role: int | None):
+    """Convert a LectureRole to text in a way that works for Reflex."""
+    return rx.cond(
+        role is None,
```
luator: Is the explicit None check needed here?  I'd expect that the default case of `rx.match` covers it.

aitutor/pages/my_lectures/components.py
```diff
@@ -0,0 +1,210 @@
...
+    return rx.text(
+        label,
+        on_click=MyLecturesState.set_role_filter(value),
+        _hover={"cursor": "pointer"},
+    )
+
+
+def role_filter_value(role: int | None):
+    """Return the filter value for a role."""
+    return rx.cond(
```
luator: I think `rx.match` could be used here instead of the nested cond.

aitutor/pages/my_lectures/components.py
```diff
@@ -0,0 +1,210 @@
...
+                max_width="100%",
+            ),
+            role_filter_button(LS.all, "all"),
+            role_filter_button(LS.owner_role, "owner"),
+            role_filter_button(LS.tutor_role, "tutor"),
+            role_filter_button(LS.student_role, "student"),
+            rx.cond(
+                MyLecturesState.is_global_admin,
+                role_filter_button(LS.not_joined, "not_joined"),
+            ),
```
luator: I think a drop-down menu would be nicer as it takes less space.

aitutor/pages/navbar.py
```diff
@@ -60,17 +60,33 @@ def navbar_link(text: str, url: str, route_to_highlight: Optional[str]) -> rx.Co
     # navigate to any admin settings page to show the admin settings navbar
     (LanguageState.admin_settings_link, routes.MANAGE_EXERCISES, "shield-check"),
 ]
+lecture_links = [
+    (LanguageState.lectures_link, routes.MY_LECTURES, "graduation-cap"),
```
luator: I'd say we could simply add this to `general_link` for easier testing

aitutor/aitutor.py
```diff
@@ -43,6 +43,26 @@
...
+app.add_page(
+    pages.all_lectures_page,
+    route=routes.ALL_LECTURES + "/[lecture_id]",
+    on_load=pages.AllLecturesState.on_load,
+)
+app.add_page(
+    pages.edit_lecture_page,
+    route=routes.EDIT_LECTURE + "/[lecture_id]",
+    on_load=pages.EditLectureState.on_load,
+)
```
luator, line 65: For some odd reason, pages with dynamic routes must be listed before pages without (no idea why but they say this in the reflex docs)

aitutor/pages/my_lectures/state.py
```diff
@@ -0,0 +1,148 @@
...
+        self.load_joined_lectures()
+
+    def on_logout(self):
+        """Clear page-specific state on logout."""
+        self._reset_filters()
+
+    @rx.var(initial_value=False)
+    def is_global_admin(self) -> bool:
+        """Whether the current user is a global admin."""
+        return GlobalPermission.ADMIN in self.global_permissions
```
luator: Since we likely will need this on other pages as well, I'd suggest to move it to the parent class (`SessionState`).

aitutor/models.py
```diff
@@ -50,6 +50,53 @@ class GlobalPermission(StrEnum):
...
+
+class Lecture(SQLModel, table=True):
+    """Lecture model for storing lecture-specific settings."""
+
+    id: Optional[int] = Field(default=None, primary_key=True)
+    check_conversation_prompt: str = Field(nullable=False, default="")
+    lecture_information_text: str = Field(nullable=False, default="")
+    lecture_name: str = Field(nullable=False, default="", unique=True)
+    lecturer_name: str = Field(nullable=False, default="")
+    registration_code: str = Field(nullable=False, default="")
```
luator, line 69: When manually browsing the database with some db viewer tools, I think it would be nicer to have the fields ordered differently. Mainly I'd put lecture_name and lecturer_name directly after id and move the prompt to the bottom.  Then the most relevant info will be visible at the left of the table without horizontal scrolling.

luator (changes requested):

aitutor/pages/all_lectures/state.py
Andrewtuebingen: Thanks, good point! I kept both route registrations because the lecture id is optional for this page, but changed the state logic to use Reflex's dynamic route variable `self.lecture_id` instead

aitutor/pages/edit_lecture/page.py
Andrewtuebingen: yes, in the beginning I thought there is only 1 owner which is the lecturer. if he loses his privileges of being a lecturer, then he shouldn't be able to edit it. but yes I will change this in the future

aitutor/pages/edit_lecture/state.py
Andrewtuebingen: I went defensive but with the new implementation no. I removed it

aitutor/pages/manage_users/state.py
Andrewtuebingen: yes in the beginning I thought there could be only 1 owner and should only be the creator (so no assigning is possible). so in the future I will change the logic due to multiple owner implementation. But yes I like the assigning idea, do you want me to do the change in this PR? because I can do it in the next one since I already have to implement role assigning in lecture-specific settings

aitutor/pages/edit_lecture/state.py
Andrewtuebingen: Good question. We do store model objects in state in some list/read views, but I think this edit form is different. These fields are editable form state with unsaved changes, so keeping them as separate primitive fields makes the form logic clearer and avoids directly changing the loaded `Lecture` object before saving.

aitutor/pages/manage_users/state.py
luator: I wouldn't do anything for directly assigning another owner here (that can be done manually once the lecture member management is implemented) but simply reject deleting of the user with an error in case they are owner of at least one lecture that does not have another owner.

aitutor/pages/my_lectures/components.py
Andrewtuebingen: good idea!

luator (approved):
Thanks a lot.

Suggestion: Let's always use the cards for the lecture list, also on desktop. I think they look more modern, are more flexible design-wise and we have to maintain less code that way.
Since that is a trivial change, I'm already approving, so you can merge without further review.

Andrewtuebingen:
> Thanks a lot.
> 
> Suggestion: Let's always use the cards for the lecture list, also on desktop. I think they look more modern, are more flexible design-wise and we have to maintain less code that way. Since that is a trivial change, I'm already approving, so you can merge without further review.

yes I will implement this in the new branch

PR #345, opened 2026-05-06
More fine-grained permission checks for displaying navbar links

luator:
I noticed that for users with the "maintainer" permission (but who are not admin or tutor), the submissions page is linked in the navbar, even though the user does not have permission to access it.  So I refactored the logic that checks which links to show, to fix that and hopefully make it a bit more flexible.

- Do not group links but have a separate permission check for each of  them.  This adds some overhead but makes it more flexible.  Also replace the tuple with a dataclass to make the code easier to  understand.
- Add function `has_permission` to more easily check if a user has a  specific permission.
- Remove the check for the MAINTAINER permission from  `lecture_has_role_at_least`.  Use the new `has_permission` function  for maintainer pages.
- Globally ignore the E711 rule (use `==`/`!=` for None) as we often need this for both `rx.cond` and SQLAlchemy.

### DIFF
--- a/aitutor/auth/protection.py
+++ b/aitutor/auth/protection.py
@@ -12,16 +12,24 @@ from aitutor.auth.state import SessionState
 from aitutor.models import GlobalPermission, UserRole
 
 
-def lecture_has_role_at_least(role):
+def lecture_has_role_at_least(role) -> rx.vars.Var[bool]:
+    """Check if the user has the specified role.
+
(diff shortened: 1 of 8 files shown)


luator:
I think there will be some conflicts with #343, so I'll wait until the other PR is merged and then resolve the conflicts here.

Andrewtuebingen:
great, thanks a lot!

Andrewtuebingen (approved):

PR #351, opened 2026-05-22
Add SMTP account email support

### LINKED ISSUE #349: Add SMTP email support for account emails
It would be useful if AI Tutor could send account-related emails, starting with signup confirmation emails and later password reset emails. Right now registration stores an email address, but the app does not seem to have SMTP/email sending configured.

This issue is meant to add the reusable SMTP setup first, so later account flows do not each invent their own mail handling.

Suggested scope:

- Add SMTP configuration via environment variables, including host, port, sender address, optional username/password, and TLS/SSL settings.
- Add a small internal helper for sending plain text emails.
- Make missing SMTP configuration fail clearly when email sending is attempted.
- Document the environment variables in the setup/deployment docs.
- Add tests for configuration parsing and message construction without sending real email.

Out of scope for this issue:

- Full password reset flow.
- Deciding the final signup confirmation UX.
- Sending real emails from tests.

Acceptance criteria:

- A developer can configure SMTP in `.env`/deployment environment and send account emails through one shared helper.
- Invalid or missing required SMTP settings produce a clear error.
- Tests cover the mail helper without contacting an SMTP server.
- Documentation explains the required and optional SMTP settings.

Quality gates:

- `pytest` passes.
- No credentials are committed.
- The implementation uses the standard config/error handling patterns already used in the project.

### LINKED ISSUE #350: Add password reset flow
Users can currently change their password after logging in, but there is no way to recover access if they forget their password or username. Once SMTP/email support exists https://github.com/martius-lab/ai-tutor/issues/349, AI Tutor should support a secure password reset flow.

Suggested flow:

- Add a "forgot password" link on the login page.
- Let users request a reset by username or email.
- Always show a neutral success message so the page does not reveal whether an account exists.
- Send a time-limited reset link by email.
- Let the user set a new password from that link.
- Mark reset tokens as used after a successful reset.
- Invalidate existing sessions after the password changes.

Acceptance criteria:

- A user who controls the account email can reset their password without admin help.
- Reset links expire and cannot be reused.
- The request form does not reveal whether a username/email exists.
- Passwords continue to be stored only as hashes.
- The flow has tests for success, expired token, reused token, and unknown account request.

Quality gates:

- `pytest` passes.
- No reset token is stored in plaintext.
- No real email is sent during automated tests.
- The implementation builds on the shared SMTP/email helper instead of adding separate mail code.

SebastianBoehler:
## Summary

Adds SMTP-backed account email support for signup welcome messages.

Closes #349.

## What changed

- Add `aitutor.mail` with SMTP settings loaded from environment variables.
- Send a localized signup welcome email with the chosen username and login URL after successful registration.
- Store the selected registration language on the new `UserInfo` row so account emails use the user's selected language.
- Document SMTP and public URL settings, including an optional Mailpit setup for local SMTP testing.
- Cover the SMTP helper and signup welcome email flow with tests that do not send real email.

## Split follow-up

Password reset was split out after review feedback and is now in #368.

## Quality gates

- `uv run ruff check aitutor/account_emails.py aitutor/mail.py aitutor/auth/state.py aitutor/pages/login_and_registration/components.py aitutor/pages/login_and_registration/state.py aitutor/language_state.py tests/test_account_emails.py tests/test_mail.py`
- `uv run pytest tests/test_mail.py tests/test_account_emails.py`
- `docker compose -f compose.yaml -f compose.mailpit.yaml config`
- `COMPOSE_BAKE=true docker compose -f compose.yaml -f compose.mailpit.yaml build app`
- Manual Mailpit check: sent signup welcome emails through `aitutor.account_emails.send_signup_welcome_email`, including from the built app image on the compose network with `SMTP_HOST=mailpit`; Mailpit captured the messages in its API and UI.

GitHub checks are passing: `pytest`, `pyright`, `ruff-check`, `ruff-format`, and `docker-build`.


SebastianBoehler:
Addressed the localization issue: account email/reset UI labels now use `LanguageState`, event/status messages use `BackendTranslations`, and welcome/password-reset email subject/body text is generated from the user/session language. I also added German/English coverage for the account email helpers and moved a module-level pages import in `SessionState` into `perform_logout()` to avoid the circular import exposed by importing translations in tests.

luator:
Thanks for tackling the email support.  This PR is quite large, though, which makes review a bit difficult.  If feasible, please split it and move the password reset stuff to a separate PR.  From what I saw, it's a separate commit, so hopefully, just moving this commit to a new PR would make sense?

SebastianBoehler:
Split this up now.

#351 is reduced to SMTP setup + signup welcome emails only. The password reset flow moved to #368.

I also updated the PR description so it no longer claims to close #350. The reset PR is currently stacked on top of the SMTP helper, so it may show some overlap until #351 lands; I can rebase it after that.

Also, do you have a rough timeline for #355/#357? A few pending issues/features seem to overlap with the lecture/exercise work, so knowing whether those are close to merge would help decide what still needs separate follow-up.

aitutor/pages/login_and_registration/state.py
```diff
@@ -100,14 +109,30 @@ def handle_custom_registration(self, form_data):
...
+            welcome_email_sent = False
+            welcome_email_failed = False
             with rx.session() as session:
-                session.add(
-                    UserInfo(
-                        email=form_data["email"],
-                        role=UserRole.STUDENT,
-                        user_id=self.new_user_id,
-                    )
+                local_user = session.get(LocalUser, self.new_user_id)
```
luator: For better readability, please move this line down to where `local_user` is actually used.

aitutor/auth/state.py
```diff
@@ -75,6 +74,8 @@ async def perform_logout(self):
         """
         Handles the logout process for the authenticated user.
         """
+        from aitutor import pages
```
luator, line 77: What's the reason for moving this down here?

aitutor/account_emails.py
```diff
@@ -0,0 +1,34 @@
...
+from reflex_local_auth.user import LocalUser
+
+from aitutor.mail import send_text_email
+from aitutor.models import UserInfo
+
+
+def public_base_url() -> str:
+    """Public base URL used for links in account emails."""
+    return cast(
+        str, config("AITUTOR_PUBLIC_URL", default="http://localhost:3000", cast=str)
```
luator: We already use a env variable `DOMAIN`, which contains the domain but without the "http"-part.  We could use that and simply assume https (one would have to manually adjust it then, if testing locally, but I think that's fine).

aitutor/language_state.py
```diff
@@ -1518,6 +1542,37 @@ def invalid_password(language: Language) -> str:
...
+                f"Sie können sich hier anmelden:\n{login_url}\n\n"
+                "Falls Sie dieses Konto nicht erstellt haben, kontaktieren Sie bitte das AI Tutor Team."
+            ),
+            en=(
+                f"Hello {username},\n\n"
+                "Your AI Tutor account has been created successfully.\n\n"
+                f"Username: {username}\n\n"
+                f"You can log in here:\n{login_url}\n\n"
+                "If you did not create this account, please contact the AI Tutor team."
+            ),
```
luator: Using a multi-line string (`""" .... """`) would make this easier to read.  You can combine it with `textwrap.dedent` so you can keep proper indentation in the code:
```python
   de=textwrap.dedent(f"""
       Hallo {username},
       
       Ihr AI Tutor Konto...
   """).lstrip()
```

docs/docker.md
```diff
@@ -20,6 +20,43 @@ Then open [https://localhost](https://localhost) in your browser to access the a
...
+
+```
+SMTP_HOST=smtpserv.uni-tuebingen.de
+SMTP_PORT=587
+SMTP_USE_TLS=true
+```
+
+The SMTP username may be the ZDV-ID rather than the visible email address.  In
+that case, set `SMTP_USERNAME` to the ZDV-ID and set `SMTP_FROM_EMAIL` to the
+actual sender address that should appear on account emails.
```
luator: While useful for us, I'd rather not add information specific to the university to the documentation.

luator (changes requested):
Thanks!  Looks mostly good, just a few comments on some details.
Also special thanks for adding unit tests :)

SebastianBoehler:
Addressed the review comments.

- Moved `local_user` down to where it is used.
- Switched welcome-email login links to use the existing `DOMAIN` env variable and assume HTTPS.
- Reworked the welcome email bodies to use `textwrap.dedent` with multi-line strings.
- Removed the Uni-specific SMTP details from the Docker docs.
- Updated the tests for the `DOMAIN` change.

I also tried moving `from aitutor import pages` back to module scope, but that reintroduced the circular import during `tests/test_account_emails.py` collection (`language_state -> auth.state -> pages -> auth.protection -> auth.state`). So I kept that import inside `perform_logout()`, where it avoids the cycle.

luator:
> I also tried moving `from aitutor import pages` back to module scope, but that reintroduced the circular import during `tests/test_account_emails.py` collection (`language_state -> auth.state -> pages -> auth.protection -> auth.state`). So I kept that import inside `perform_logout()`, where it avoids the cycle.

I wonder if there is a better way to resolve the cyclic import but let's keep it like that for now (we should anyway reconsider our current usage of `pages/__init__.py`).

luator:
I tested it now and noticed some issues:  I configured SMTP data and registered an account.  Sending the mail failed due to timeout.  This might be a mistake in the config (I didn't check yet), but it revealed the following issues:

- Sending the mail should be done asynchronously. At least, there should be some feedback to the user that something is happening.  Currently, after clicking the "Register" button, nothing happens (i.e. it seems like one did not properly click the button) for 10 seconds until the timeout is triggerd.
- After that, the "welcome_email_failed" callout is shown for a short moment, but then the page is immediately redirected to the login page, so I could not even read the callout message.
- Failures of sending emails should be printed to the log (i.e. stdout), so we can actually know about it and debug it.  I simply added a print in the `except` block of `send_mail()` for that.  Not sure if there is a better way.

aitutor/mail.py
```diff
@@ -0,0 +1,138 @@
...
+    client_cls = smtp_client_cls or (SMTP_SSL if active_settings.use_ssl else SMTP)
+
+    try:
+        with client_cls(
+            active_settings.host,
+            active_settings.port,
+            timeout=active_settings.timeout,
+        ) as smtp:
+            if active_settings.use_tls:
+                smtp.starttls()
```
luator, line 131:
```suggestion
                smtp.starttls()
                smtp.ehlo()
```
According to the documentation, `ehlo()` should be called after `starttls()`.  Adding this fixed an error for me (without, sending of the mail failed with the error "[SSL: WRONG_VERSION_NUMBER] wrong version number").

SebastianBoehler:
Thanks, I pushed the follow-up changes and resolved the addressed thread.

- Added `smtp.ehlo()` immediately after `smtp.starttls()`.
- Registration now shows a loading state while the welcome email is being sent.
- The welcome email send runs in a worker thread, so SMTP does not block the backend event loop.
- SMTP configuration/delivery failures are logged.
- If the welcome email fails, the account is still created, but the page keeps the warning callout visible instead of redirecting immediately.
- The branch is updated against current `main`.

aitutor/pages/login_and_registration/state.py
```diff
@@ -91,24 +97,44 @@ def handle_custom_registration(self, form_data):
...
-            return
-
-        # check for the correct registration code
-        registration_code = get_config().registration_code
-        if form_data["registration_code"] != registration_code:
-            self.error_message = "The registration code is wrong."
-            self.registration_code = ""
-            return
-
-        registration_result = self.handle_registration(form_data)
```
luator, line 101: What is the reason not using `handle_registration` anymore?  Sticking to that seems more future-proof to me than replicating its functionality and calling private methods of the parent class.  Their behaviour might change for new versions of the reflex-local-auth package, thus breaking our code.

aitutor/account_emails.py
```diff
@@ -17,17 +17,27 @@ def public_base_url() -> str:
 
 def send_signup_welcome(*, local_user: LocalUser, user_info: UserInfo) -> None:
```
luator: Not used anymore?

If I understood correctly, you are mostly ai-generating your code?  It would be great if you could review it a bit more closely before submitting.

SebastianBoehler:
Thanks, both points are fair.

Yes, the code is AI-generated, but I am reviewing the changes before submitting. In this case I missed the point in the latest iteration and left some issues in the diff. I’ve addressed them now.

Sorry for the noise. I am still iterating on my setup for this repo. The repo is not currently set up for coding-agent workflows, and Reflex also has some unusual patterns, which makes it less straightforward than my own projects. Since I do not know whether you use coding agents here, I held back from adding agent-specific files like `AGENTS.md` or repo-specific skills. Those would improve consistency and code quality for AI-assisted changes, but I did not want to add that kind of setup when I would be the only one benefiting.

I would love to set this repo up for coding agents or create a PR so you could see how that would look like bc then PRs literally get direct merges as code quality is incredibly good when used properly, thats also what I am used to so I might trust it too much in this repo

I’ll review future PR updates more closely before pushing.

aitutor/mail.py
```diff
@@ -32,20 +32,26 @@ class SmtpSettings:
...
+        default_port = 465 if use_ssl else 587
         settings = cls(
-            host=cast(str, config("SMTP_HOST", default="", cast=str)).strip(),
-            port=cast(
-                int, config("SMTP_PORT", default=465 if use_ssl else 587, cast=int)
+            host=app_settings.smtp_host,
+            port=(
+                app_settings.smtp_port
+                if app_settings.smtp_port is not None
+                else default_port
```
luator: Assuming that 0 is not a valid port, this could be simplified to `port=app_settings.smtp_port or default_port`.

aitutor/mail.py
```diff
@@ -32,20 +32,26 @@ class SmtpSettings:
...
+            ),
+            from_email=app_settings.smtp_from_email,
+            username=app_settings.smtp_username,
+            password=app_settings.smtp_password,
+            use_tls=(
+                not use_ssl
+                if app_settings.smtp_use_tls is None
+                else app_settings.smtp_use_tls
```
luator: I find this a bit confusing.  Defaults to TLS unless SSL is enabled or TLS is explicitly disabled?  I'd keep it simple and have both default to false (so we explicitly need to enable the one we want in the config).

luator:
I'm not using coding agents but I don't know about @Andrewtuebingen.  I wouldn't mind adding an AGENTS.md as I can simply ignore it if I don't need it.  I'd actually be curious to see how it changes the code quality.  Right now the changes provided by you look mostly good but contain some rather obvious AI quirks that make reviewing a bit tiresome to be honest (especially since I'm not always sure if I'm actually talking to a human or directly to the LLM).

Reflex indeed does some rather shady black magic stuff, which I can imagine being difficult for AI tools.  I frequently struggle myself with it.  As well as static type checking tools and linters who complain about the obviously wrong code that still works for whatever reason...

aitutor/app_settings.py
```diff
@@ -0,0 +1,55 @@
...
+        str, StringConstraints(strip_whitespace=True, min_length=1)
+    ] = Field(validation_alias="OPENAI_API_KEY")
+    openai_base_url: str | None = Field(
+        default=None, validation_alias="OPENAI_BASE_URL"
+    )
+    domain: Annotated[str, StringConstraints(strip_whitespace=True)] = Field(
+        default="localhost", validation_alias="DOMAIN"
+    )
+    smtp_host: Annotated[str, StringConstraints(strip_whitespace=True)] = Field(
+        default="", validation_alias="SMTP_HOST"
```
luator: Would it make sense to change the default of these fields to `None`?  Then type-checkers should force use to check if the field is actually set or not before using its value.

aitutor/mail.py
SebastianBoehler: Agreed, that is simpler. I changed this to `port=app_settings.smtp_port or default_port`.

SebastianBoehler:
Addressed the comments. Comments / replies are mainly from me. I use the agent to draft or send them via Codex directly, but comments are always from me, not necessarily writing myself, but they are iterated on by me.

May I ask why you're not making use of agents?

luator:
> May I ask why you're not making use of agents?

I use it from time to time for writing some small self-contained scripts (like for plotting some data or simple tools), stuff which I won't need to further maintain in the future.  For bigger projects I like to rather make design decisions by myself and keep an understanding of the code base.  Finally, using an Agent means replacing writing code with reviewing code.   Writing code is fun and a skill I don't want to unlearn.  Just reviewing huge amounts of code all day is not fun and quite exhaustive.
Also, based on my (admittedly somewhat limited) experience, I don't buy the claim of a huge productivity increase but it does burn quite a lot of energy and money.

aitutor/app_settings.py
```diff
@@ -25,18 +25,24 @@ class AppSettings(BaseSettings):
     domain: Annotated[str, StringConstraints(strip_whitespace=True)] = Field(
         default="localhost", validation_alias="DOMAIN"
     )
-    smtp_host: Annotated[str, StringConstraints(strip_whitespace=True)] = Field(
-        default="", validation_alias="SMTP_HOST"
+    smtp_host: Annotated[str, StringConstraints(strip_whitespace=True)] | None = Field(
+        default=None, validation_alias="SMTP_HOST"
```
luator, line 29: Sorry, there is something I missed earlier: Since sending emails isn't optional, the corresponding settings should actually be required as well.  Then ideally the validation should happen during initialisation of the app, so we fail early if something is wrong.

luator:
Side note: I found that there are various tools to set up a local "fake" SMTP server for testing.  I tried [mailpit](https://mailpit.axllent.org) now but there are others with similar functions (maildev, smtp4dev, ...).

SebastianBoehler:
Good idea. I added a small optional Mailpit compose override for local SMTP testing and documented the .env values for routing signup emails to it.

I also tested it locally: Mailpit starts via the override, the app email helper can send to it, and the built app image can reach it on the compose network via SMTP_HOST=mailpit. Mailpit captured the welcome emails in both the API and UI.

luator:
Sorry, I was quite busy with other things so I neglected this PR a bit.
I think the only bit missing here to get it merged is proper handling of missing SMTP config.  I thought about it again.  To not disrupt local development too much, I'd keep the config optional for now and if no SMTP_HOST is set, simply print the mail to stdout instead of sending it.

luator:
Work on this is continued in #394

PR #353, opened 2026-05-22
fix(auth): add browser autofill hints

### LINKED ISSUE #352: Remember last used username on login
## What

It would be useful if the login page could remember the last username a user chose to save, so returning users do not have to type it again every time.

This should only store the username, not the password or any session token. A small "Remember me" checkbox on the login form would be enough.

## Acceptance criteria

- The login form has a clear "Remember me" checkbox.
- When it is checked and login is submitted, the entered username is stored locally in the browser.
- When the login page is opened again, the stored username is pre-filled.
- When it is unchecked and login is submitted, any previously stored username is cleared.
- No password, token, or other sensitive auth data is stored.

## Quality gates

- Existing tests still pass.
- The change works after a page refresh.
- The UI remains usable on small screens.

SebastianBoehler:
## What

Adds explicit browser/password-manager autocomplete hints to the auth form fields:

- login username: `username`
- login password: `current-password`
- registration username: `username`
- registration email: `email`
- registration password fields: `new-password`

No username, password, token, or session data is stored by the app.

## Notes

The hints are passed through `custom_attrs={"auto_complete": ...}`. In the installed Reflex version, `rx.input(auto_complete=...)` is typed as a boolean and renders only `autoComplete={true}` / `autoComplete={false}`, so it cannot express field-purpose tokens like `username` or `current-password`.

Using the snake-case `auto_complete` custom attribute keeps the Python code consistent with the codebase while compiling to React's expected `autoComplete` prop.

## Validation

- `uv run ruff check aitutor/pages/login_and_registration/components.py`
- `uv run pyright aitutor/pages/login_and_registration/components.py`
- `git diff --check`
- Isolated Reflex render check: `auto_complete=True` renders `autoComplete:true`; `custom_attrs={"auto_complete": "username"}` renders `autoComplete:"username"`.
- `uv run reflex run --frontend-port 3002 --backend-port 8002`
- Browser check on `http://localhost:3002/login`: login form rendered and the React console no longer reports the invalid lower-case `autocomplete` prop warning.


luator:
Sorry for being so slow responding to your PRs, I've been out of office for a while.  I'm trying to go through them one by one now.

Regarding this one: To be honest, I don't really see the point of this feature.  If I understand it correctly, all it does is pre-filling the username field of the login form?  That doesn't seem useful enough to me to justify the added code complexity.  The login session is already somewhat persistent (i.e. one doesn't immediately get logged out when closing the browser).  If someone wants to store the login information, I think using the functionality provided by the browser (which can also store the password) would be more useful.

SebastianBoehler:
Fair point, I agree this probably does not justify the added state/localStorage path if the feature is basically only pre-filling the username.

The reason I looked into it was more practical: Chrome/password-manager autofill was not very reliable for me here. During registration it also did not consistently offer to save the credentials, and on the login page I sometimes got suggestions from a different account/domain instead. So this PR was more of a workaround for that friction than a feature with much standalone value.

I had another quick look at the form markup. The fields already have reasonable `name`s and the password input is a password field, but we do not set explicit `autocomplete` hints. A smaller follow-up might make more sense there, e.g. `autocomplete="username"` + `autocomplete="current-password"` for login, and `username` / `email` / `new-password` for registration. That is also closer to the browser/password-manager path you suggested: https://web.dev/articles/sign-in-form-best-practices

So yes, I am fine closing this PR if you think the current approach is not worth it. If the autofill issue is reproducible for others, I would rather open a tiny PR for the input hints instead.

SebastianBoehler:
Updated the PR along those lines now.

I removed the localStorage/"Remember me" implementation and replaced it with the smaller browser-native version: explicit `autocomplete` hints for login and registration fields. This should work better with browser/password-manager autofill.

luator (commented):
Thanks for looking into it.  I didn't know about the `autocomplete` attribute, makes total sense to add it to the fields.

One minor request: Could you please change the attribute name to all lower case?  I think functionally it doesn't matter (HTML is case-insensitive), but it would be more consistent with our code base.

luator:
Btw, did you understand what exactly the `auto_complete` attribute of `rx.input` does?  I thought it might make sense to additionally set this to True but when I tested it, I couldn't see any difference in the produced HTML code.

luator:
(and a side note: As is, the pull request description is confusing as it is referring to the previous content, which doesn't exist anymore)

SebastianBoehler:
Updated, thanks. I changed the code-side attribute key from `autoComplete` to snake-case.

I also looked into `rx.input(auto_complete=...)`. `TextFieldRoot.auto_complete` is declared as a boolean prop: an isolated render check shows `auto_complete=True` emits `autoComplete:true`, while passing string tokens directly is rejected. The token-bearing path is therefore still `custom_attrs`; using `custom_attrs={"auto_complete": "username"}` renders `autoComplete:"username"` in the generated Reflex route. I briefly tried the literal lower-case `autocomplete` key too, but React logs an invalid DOM property warning for that, so the snake-case custom attr is the cleaner lower-case Python spelling.

And updated pr description

luator:
Why snake case?  The HTML attribute is "autocomplete".  I assume changing it to "auto_complete" doesn't work (I'm taking about the key in `custom_attrs`, not the attribute of `rx.input`).

SebastianBoehler:
Good point. I verified the actual DOM now: `custom_attrs={"auto_complete": "username"}` renders as HTML `autocomplete="username"` on the input. Reflex converts the Python snake-case key to React’s `autoComplete` prop, and React writes the lower-case HTML attribute.

I also tested the literal `custom_attrs={"autocomplete": ...}` key, but React logs an invalid DOM property warning for that and suggests `autoComplete`, so I’d keep the snake-case Reflex spelling.

luator:
Ah, okay.  I assumed that stuff from `custom_attr` is directly added to the HTML (reflex docs are sadly lacking on this one...).  But indeed, I tested and `auto_complete` gets translated to `autocomplete`.

luator (approved):

PR #355, opened 2026-05-22
Andrew/home lecture

### LINKED ISSUE #311: Req 10: Users can leave lectures they previously joined
---
See #301

### LINKED ISSUE #314: Req 19: In-lecture role "Owner" can add users to the lecture
---
See #301

### LINKED ISSUE #315: Req 20: In-lecture role "Owner" can assign roles to users within the lecture
Available roles are "Student", "Tutor" and "Owner".

---
See #301

### LINKED ISSUE #316: Req 23: A lecture must always have at least one "Owner"
That means

1. One cannot change the role of the last owner.
2. A sole owner cannot leave the lecture (has to assign the owner role to another user first, see #315).

---
See #301

### LINKED ISSUE #341: Lecture overview
- leave navigation above after entering a lecture
- Lecture = my_lecture
- remove navigation of edit_lecture and all_lecture
- make the overview lecture specific

### LINKED ISSUE #346: multiple owner implementation
In the beginning I thought there is only 1 owner which is the lecturer. if he loses his privileges of being a lecturer, then he shouldn't be able to edit it. since we might have multiple owners in a lecture, everybody from them should be able to edit the lecture they own.

### LINKED ISSUE #347: ensure that every lecture will always have 1 owner
- for role changing
- for kicking

Andrewtuebingen:
## Summary

This PR adds and refactors lecture related functionality around lecture overview, lecture members, permissions, and lecture management.

### Added

- Added a dedicated lecture overview page with lecture-specific navigation.
- Added a lecture members page for managing members of a lecture.
- Added role management for lecture members.
- Added functionality to add users as lecture members.
- Added functionality for members to leave a lecture.
- Added functionality to kick members from a lecture.
- Added a warning/confirmation step before kicking a member.
- Added lecture-specific navbar/navigation components.
- Added shared lecture permission utilities to centralize permission checks.


aitutor/pages/edit_lecture/state.py
```diff
@@ -269,19 +280,22 @@ def _normalized_lecturer_name(self) -> str:
         """Return the trimmed lecturer name used for validation and persistence."""
         return self.lecturer_name.strip()
 
+    def _user_may_create_lecture(self) -> bool:
+        """Check whether the current user may create a new lecture."""
+        return (
+            GlobalPermission.LECTURER in self.global_permissions
+            or GlobalPermission.ADMIN in self.global_permissions
+        )
```
luator: Since we're going to need this kind of check a lot in various places, I suggest that we add a function `has_permission()` to SessionState, which takes the permission as input (LECTURER) in this case and internally also checks for ADMIN.

aitutor/pages/lecture_members/state.py
```diff
@@ -0,0 +1,391 @@
...
+            LectureMemberRow(
+                user_id=member.user_id,
+                username=member.username,
+                role=member.role,
+                selected_role=role_name
+                if member.user_id == user_id
+                else member.selected_role,
+            )
+            for member in self.members
+        ]
```
luator: Rewriting all the members for changing only one of them seems a bit dangerous to me.
How about changing `self.members` to a dictionary "user_id -> LectureMemberRow"?  Then you should be able to simply do
```python
self.members[user_id].selected_role = role_name
```
here.  If you need them as list somewhere, that would still be possible with `list(self.members.values())`.

aitutor/pages/lecture_members/state.py
```diff
@@ -0,0 +1,391 @@
...
+                ).all()
+            )
+
+            users = session.exec(select(LocalUser).order_by(LocalUser.username)).all()
+
+        self.available_users = [
+            AvailableLectureUser(user_id=int(user.id), username=user.username)
+            for user in users
+            if user.id is not None and user.id not in member_user_ids
+        ]
```
luator, line 326: I think this would be better done in a database query.  I think the following SQL should work (untested, though, so take with a grain of salt!):
```sql
SELECT * FROM localuser
WHERE localuser.user_id NOT IN (SELECT user_id FROM linkuserlecture WHERE lecture_id = ...)
ORDER BY localuser.username
```
That should hopefully be possible to translate to  SQLAlchemy without much difficulty.

aitutor/pages/lecture_overview/state.py
```diff
@@ -0,0 +1,75 @@
...
+        self.lecture_name = ""
+        self.lecturer_name = ""
+        self.lecture_information_text = ""
+
+    def _apply_lecture_to_state(self, lecture: Lecture) -> None:
+        """Copy the loaded lecture into state variables for rendering."""
+        self.current_lecture_id = lecture.id
+        self.lecture_name = lecture.lecture_name
+        self.lecturer_name = lecture.lecturer_name
+        self.lecture_information_text = lecture.lecture_information_text
```
luator, line 58: The data here is only used for displaying, right?  Would it be possible to simply store the `lecture` object in the state?  (I think I had a similar question already recently, so sorry if I forgot a reason why this doesn't work).

aitutor/pages/lecture_overview/state.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def _get_route_lecture_id_param(self) -> str:
+        """Return the lecture id route parameter."""
+        return str(self.lecture_id)
+
+    def _user_may_view_lecture(self, lecture_id: int) -> bool:
+        """Check whether the current user may open this lecture."""
+        if self.authenticated_user is None or self.authenticated_user.id is None:
+            return False
```
luator, line 67: Not really related to this PR, but I think we have this check quite often in our code.  It might make sense to move it to `SessionState` and call it something like `is_user_logged_in()`.  But maybe makes more sense to do this in a separate PR and then replace everywhere.

aitutor/pages/my_lectures/components.py
```diff
@@ -103,12 +106,18 @@ def role_filter_text(label, value) -> rx.Component:
...
+        ROLE_FILTER_OWNER,
+        rx.cond(
+            role == LectureRole.TUTOR.value,
+            ROLE_FILTER_TUTOR,
+            rx.cond(
+                role == LectureRole.STUDENT.value,
+                ROLE_FILTER_STUDENT,
+                ROLE_FILTER_NOT_JOINED,
+            ),
+        ),
```
luator: Why not keeping the `rx.match`?  The new code seems to be functionally the same.  Maybe I missed something?

aitutor/pages/my_lectures/components.py
```diff
@@ -140,6 +149,98 @@ def browse_lectures_button() -> rx.Component:
...
 
 
+def leave_lecture_button(lecture: Lecture, *, width: str | None = None) -> rx.Component:
+    """Render the leave lecture button with a destructive confirmation dialog."""
+    return destructive_confirm(
+        title=LS.leave_lecture + ": " + lecture.lecture_name,
+        description=LS.leave_lecture_description,
+        confirm_text=LS.leave_lecture,
+        cancel_text=LS.cancel,
+        on_confirm=MyLecturesState.leave_lecture(cast(int, lecture.id)),
```
luator: I assume the `cast` is needed as `lecture.id` is `int | None`?  It might be better to be sure and explicitly check, e.g. by adding `assert lecture.id is not None` (I hope the type checker will understand that, not sure).
Then, in case we accidentally pass a bad lecture object here, we would get a clear error message.

aitutor/pages/navbar_specific_lecture.py
```diff
@@ -0,0 +1,105 @@
+"""Specific lecture navbar."""
```
luator, line 1: I think the filename a bit confusing.  Shouldn't it rather be `lecture_specific_navbar.py`?  Actually, I think `lecture_navbar.py` should be enough.

aitutor/pages/navbar_specific_lecture.py
```diff
@@ -0,0 +1,105 @@
...
+
+import reflex as rx
+
+import aitutor.routes as routes
+from aitutor.language_state import LanguageState
+
+specific_lecture_links = [
```
luator, line 21: For better readability of the code, I'd rather use a [named tuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple).  Then you can use field names instead of (easy to confuse) indices in the code below.

aitutor/pages/my_lectures/state.py
```diff
@@ -41,6 +43,53 @@ def set_role_filter(self, value: str):
...
+
+        user_id = self.authenticated_user.id
+
+        with rx.session() as session:
+            link = session.exec(
+                select(LinkUserLecture).where(
+                    LinkUserLecture.lecture_id == lecture_id,
+                    LinkUserLecture.user_id == user_id,
+                )
+            ).one_or_none()
```
luator: Use `get_user_lecture_link()` here?

aitutor/utilities/lecture_permissions.py
```diff
@@ -0,0 +1,92 @@
...
+    """Return the user's lecture role, if they are a member of the lecture."""
+    link = get_user_lecture_link(session, user_id=user_id, lecture_id=lecture_id)
+    return link.role if link is not None else None
+
+
+def has_global_admin_permission(
+    global_permissions: Collection[GlobalPermission],
+) -> bool:
+    """Return whether the user has global admin permissions."""
+    return GlobalPermission.ADMIN in global_permissions
```
luator, line 40: Depending on where/how exactly this is used, it might be redundant with `SessionState.is_global_admin`.

aitutor/utilities/lecture_permissions.py
```diff
@@ -0,0 +1,92 @@
...
+        return True
+
+    return (
+        get_user_lecture_role(
+            session,
+            user_id=user_id,
+            lecture_id=lecture_id,
+        )
+        == LectureRole.OWNER
+    )
```
luator: No need for branching here
```suggestion
    return has_global_admin_permission(global_permissions) or (
        get_user_lecture_role(
            session,
            user_id=user_id,
            lecture_id=lecture_id,
        )
        == LectureRole.OWNER
    )
```
But maybe a bit less readable, which is probably the more important concern.  I leave it to you, which version you prefer.

aitutor/pages/lecture_members/components.py
```diff
@@ -0,0 +1,293 @@
...
+        rx.dialog.content(
+            rx.vstack(
+                rx.dialog.title(LS.add_member),
+                rx.dialog.description(LS.add_member_description),
+                available_user_filter(),
+                available_users_list(),
+                rx.hstack(
+                    rx.dialog.close(
+                        rx.button(
+                            LS.cancel,
```
luator: I think it should rather say "Close" than "Cancel" here.  "Cancel" somewhat suggests that already performed "Add" actions might be undone.
Instead of a "Close" button a "x"-button in the top right might also work (I think we did this somewhere already, not sure).

aitutor/pages/navbar_specific_lecture.py
```diff
@@ -0,0 +1,105 @@
...
+    ),
+    (
+        LanguageState.exercises_link,
+        routes.LECTURE_OVERVIEW,
+        "book-open",
+        "exercises",
+        True,
+    ),
+    (LanguageState.members, routes.LECTURE_MEMBERS, "users", "members", False),
+    (LanguageState.settings, routes.LECTURE_OVERVIEW, "settings", "settings", True),
```
luator: This should point to the edit_lecture page, right?  I suggest to directly enable that and remove the "Edit" button from the my_lectures list.

luator (changes requested):
Sorry that it took me so long to do the review.  I have some comments on implementation details but functionality looks good.

Something I noticed while testing:  When creating a new lecture, after saving, I stay at the edit_lecture form.  I would prefer if upon creation, I would be redirected to the  lecture_overview page (i.e. same as if I "enter" a lecture)

aitutor/pages/navbar_specific_lecture.py
Andrewtuebingen: I used the navber prefix for consistency as that what the other navbar filename start with. also we have a navbar_lectures.py already for all_lecture, my_lecture, edit_lecture. so the navbar_specific_lecture.py is conceptually the subnavbar of navbar_lectures.py. but I understand that it sounds weird

aitutor/pages/navbar_specific_lecture.py
Andrewtuebingen: great idea

aitutor/utilities/lecture_permissions.py
Andrewtuebingen: It is the same underlying check, but I think this helper is still useful here because `user_may_view_lecture()` and `user_may_edit_lecture()` are state independent utility functions. They receive `global_permissions` explicitly and should not depend on `SessionState`. `SessionState.is_global_admin` is convenient inside state/components, but `has_global_admin_permission()` keeps these permission helpers independent from Reflex state.

aitutor/pages/navbar_specific_lecture.py
Andrewtuebingen: That's a good question because I don't know what we should label as the settings tab because I thought this might be for manage_exercise or prompt or configuration. maybe we can discuss on wednesday what would be the best solution

aitutor/utilities/lecture_permissions.py
Andrewtuebingen: I prefer your suggestion

aitutor/pages/lecture_overview/state.py
Andrewtuebingen: true, I made an issue for future pr

aitutor/pages/my_lectures/components.py
Andrewtuebingen: oh this happened accidentally, maybe while resolving a merge conflict

aitutor/pages/lecture_overview/state.py
Andrewtuebingen: Yes, here it would probably be possible because the overview only displays lecture data and does not edit it (like the other case where you asked a similar question). I kept the fields explicit to store only the data needed for rendering and avoid exposing unused lecture settings/relationships in the page state like  `registration_code`, `check_conversation_prompt`and `user_links`. But its up to you if you want me to change it

Andrewtuebingen:
Thanks for the suggestion! I implemented most of them

aitutor/pages/lecture_overview/state.py
luator: Good point, I didn't think about exposing additional fields.  Let's keep it as is.

aitutor/pages/navbar_specific_lecture.py
luator: With that context it makes more sense to me, i.e. it should be read as "the navbar for 'specific lecture'".  Previously I read it more as "navbar-specific lectures" which was confusing.
Do we actually still need/use the all_lecture/my_lecture/edit_lecture navbar?  If not, this could go away and the specific one be renamed to "navbar_lecture.py"?

luator (approved):

luator (changes requested):
As mentioned in https://github.com/martius-lab/ai-tutor/issues/359#issuecomment-4658517714, please add all new state classes in `SessionState.perform_logout`.

Andrewtuebingen:
I added all missing state classes in SessionState.perform_logout. I will fix the privacy bug in the next pr

luator (approved):

PR #357, opened 2026-05-27
Andrew/exercise

### LINKED ISSUE #317: Req 13: In-lecture role "Student" can see/work on the exercises of that lectures
---
See #301

### LINKED ISSUE #322: Req 21: In-lecture role "Owner" can create/edit/delete Exercises in that lecture

### LINKED ISSUE #323: Req 22: In-lecture role "Owner" can create/edit/delete Tags that can be assigned to exercises of that lecture.
This means tags are lecture-specific, there are no global tags.

---
See #301

Andrewtuebingen:
## Implemented lecture-specific exercise management and lecture-specific exercises

In this PR, it is not yet possible to click on an exercise and continue with it in the chat. That will be handled in a future PR.

Everything else has already been implemented in a lecture-specific way. I intentionally kept the global exercise and global manage exercise functionality for now, because I would prefer to remove  the global parts only once the full implementation is finished.

For now, it is still possible to create both global exercises and lecture-specific exercises.


aitutor/pages/lecture_exercises/state.py
```diff
@@ -0,0 +1,284 @@
...
+from aitutor.utilities.filtering_components import FilterMixin
+from aitutor.utilities.lecture_permissions import user_may_view_lecture
+
+ExerciseWithResult = tuple[Exercise, Optional[ExerciseResult]]
+
+
+class LectureExercisesState(FilterMixin, SessionState):
+    """State for managing exercises belonging to one lecture."""
+
+    current_lecture_id: int | None = None
```
luator: Do we need this as separate field?  I'd think that using `int(self.lecture_id)` where it is needed should be enough.  Duplicating the value bears the risk of getting out of sync at some point.

aitutor/pages/lecture_exercises/state.py
```diff
@@ -0,0 +1,284 @@
...
+
+        if not self._user_may_view_lecture(lecture_id):
+            return rx.redirect(routes.MY_LECTURES)
+
+        with rx.session() as session:
+            if session.get(Lecture, lecture_id) is None:
+                return rx.redirect(routes.NOT_FOUND)
+
+        self.current_lecture_id = lecture_id
+        assert self.authenticated_user_info is not None
```
luator: I think this assert should rather be at the top of the method.

alembic/versions/c9782a9331d4_.py
```diff
@@ -0,0 +1,40 @@
...
+# revision identifiers, used by Alembic.
+revision: str = 'c9782a9331d4'
+down_revision: Union[str, None] = 'df4ec81ec5d5'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
```
luator, line 29: Important: When altering tables, add the code to disable foreign key constraints for sqlite (see in existing alembic scripts).  Otherwise data may be lost due to cascaded delete (sqlite cannot alter tables, so what happens is that the old table is deleted and a new, modified one created).

luator (changes requested):
I tested and everything works as expected.
I didn't review in detail, assuming that this is mostly a copy&paste from the global exercises pages.
One important change request on the alembic file, though, see below.

aitutor/pages/lecture_exercises/state.py
Andrewtuebingen: good point, I removed the duplicated `current_lecture_id` state from `LectureExercisesState`. The lecture id is now derived from the route via `_parse_route_lecture_id()` where needed, with guards to avoid querying `Exercise.lecture_id == None`.

alembic/versions/c9782a9331d4_.py
Andrewtuebingen: great catch!

Andrewtuebingen:
Thanks for the review!

luator (approved):
Thanks!

PR #361, opened 2026-06-09
[codex] fix protected page auth flash

### LINKED ISSUE #359: privacy bug
when typing ex: http://localhost:3000/lectures/edit_lecture/4 in url, the unauthorized user can see the page for a couple of milliseconds before getting redirected

SebastianBoehler:
## Summary
- wrap protected pages with `page_require_role_or_permission` outside page/navbar decorators
- prevent protected page chrome from rendering before the existing auth guard resolves

## Root cause
The auth guard was applied inside navbar/page wrapper decorators. The page body was gated, but protected route chrome could render briefly before `LoginState.redir` settled the unauthenticated session.

## Validation
- `uv run pytest`
- `uv run ruff check aitutor/pages`
- In-app browser smoke: unauthenticated `/lectures/edit_lecture/4` redirects to `/login`; no lecture edit data was observed during testing

Fixes #359

### DIFF
--- a/aitutor/pages/all_lectures/page.py
+++ b/aitutor/pages/all_lectures/page.py
@@ -10,9 +10,9 @@ from aitutor.pages.navbar import with_navbar
 from aitutor.pages.navbar_lectures import with_lectures_navbar
 
 
+@page_require_role_or_permission(required_role=UserRole.STUDENT)
 @with_navbar(routes.LECTURES)
 @with_lectures_navbar(routes.ALL_LECTURES)
-@page_require_role_or_permission(required_role=UserRole.STUDENT)
 def all_lectures_page() -> rx.Component:
     """Show the page scaffold for all available lectures."""
     return rx.center(
--- a/aitutor/pages/chat/page.py
+++ b/aitutor/pages/chat/page.py
@@ -16,8 +16,8 @@ from aitutor.pages.chat.state import ChatState
 from aitutor.pages.navbar import with_navbar
 
 
-@with_navbar(routes.EXERCISES)
 @page_require_role_or_permission(required_role=UserRole.STUDENT)
+@with_navbar(routes.EXERCISES)
 def chat_page() -> rx.Component:
     """Renders the web page."""
     return rx.container(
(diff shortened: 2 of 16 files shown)


tests/test_protected_page_decorators.py
```diff
@@ -0,0 +1,59 @@
...
+    assert "EditLectureState.page_is_ready" in component_source
+
+
+def test_auth_hydration_fallback_is_not_visible():
+    """Protected routes should not display page data while auth hydrates."""
+    protection_source = (AUTH_DIR / "protection.py").read_text()
+
+    assert "on_mount=LoginState.redir" in protection_source
+    assert 'display="none"' in protection_source
+    assert "Loading..." not in protection_source
```
luator: Do I understand correctly, that these tests just test for the listed strings to appear somewhere in the code?  That doesn't seem very useful/reliable to me.

aitutor/pages/edit_lecture/components.py
```diff
@@ -141,82 +141,86 @@ def edit_lecture_form() -> rx.Component:
...
-                ),
-                rx.cond(
-                    EditLectureState.unsaved_changes,
-                    rx.callout(
-                        LS.unsaved_changes_info,
-                        icon="info",
-                        width="100%",
-                        color_scheme="orange",
+    return rx.cond(
+        EditLectureState.page_is_ready,
```
luator: Is this `page_is_ready`-guarding really needed?  It shouldn't expose any sensitive data if protection in `on_load()` of the state is done properly, so I'd rather keep the code simple here.

SebastianBoehler:
Pr is still in draft. Treat tests as placeholders. We should introduce tests that ensure routes are protected. Open to discussion.

Whole PR can be closed anyway if on privacy issue decided that menu bar state flashing is fine

luator:
I think the swapping of decorator order is fine.  It seems to fix the flashing issue and I don't expect it to cause any problems.  Only the other change adding the `page_is_ready` flag I would omit and also the last two tests.

aitutor/pages/edit_lecture/components.py
SebastianBoehler: afaik understand reflex doesnt really provide supports for middleware and this would be the "earliest" point to reject the request but I guess other than few ms speed up doe to loading this doesnt add anything, you're right.

luator:
@Andrewtuebingen This looks good to me and seems to fix #359.  Are you aware of any potential issue with changing the decorator order, that I might be missing?

luator (approved):

luator:
Thanks

PR #364, opened 2026-06-09
fix: reset session state on logout

### LINKED ISSUE #362: Make sure all state classes are listed in SessionState.perform_logout
For the `on_logout` method to be executed, all our states need to be listed in `SessionState.perform_logout`.
Check if this is already the case or if we added any new state classes recently that are not listed there.

### LINKED ISSUE #363: More secure data clearing on logout
Each of our state classes implements a method `on_logout` to clear data.  This is currently only executed for classes that are listed in `SessionState.perform_logout`, though, which is error prone (it's easy to add a new state class and forget listing it there).
So if possible, we should come up with a better system that doesn't rely on us manually listing all state classes.

SebastianBoehler:
## Summary
- replace the manual logout cleanup list with Reflex's native `State.reset()` handling
- remove the fragile `aitutor.pages` import from `aitutor.auth.state`
- reset `SessionState` and its substates after invalidating the local auth session

## Root cause
`SessionState.perform_logout()` manually listed page state classes whose `on_logout()` methods should be called. New page states could be missed, leaving user-specific UI state cached after logout.

Reflex already provides `State.reset()`, which resets the current state and its substates. The page states inherit from `SessionState`, so this gives us the intended framework-level cleanup path without maintaining a parallel list.

## Validation
- `uv run ruff check aitutor/auth/state.py`
- `uv run pytest`
- In-app browser smoke: created a disposable local user, logged in, logged out, confirmed protected `/exercises` redirects to `/login`, logged in again, logged out again, then deleted the disposable user

Fixes #362
Fixes #363

### DIFF
--- a/aitutor/auth/state.py
+++ b/aitutor/auth/state.py
@@ -9,7 +9,6 @@ import reflex_local_auth
 from sqlmodel import select
 
 import aitutor.routes as routes
-from aitutor import pages
 from aitutor.models import GlobalPermission, Language, Permission, UserInfo, UserRole
 
 
@@ -75,24 +74,8 @@ class SessionState(reflex_local_auth.LocalAuthState):
         """
         Handles the logout process for the authenticated user.
         """
-        states = [
-            pages.ChatState,
-            pages.HomeState,
-            pages.ExercisesState,
-            pages.FinishedViewState,
-            pages.FinishedViewTutorState,
-            pages.ManageExercisesState,
-            pages.ManageTagsState,
-            pages.SubmissionsState,
-        ]
-        for state in states:
-            # get the state
-            state_instance = await self.get_state(state)
-            # clear the state
-            state_instance.on_logout()
-
-        # logout
         self.do_logout()
+        self.reset()
         return rx.redirect(routes.HOME, replace=True)
 
     @rx.var(cache=True, initial_value=None)


luator:
Thanks, I wasn't aware of the `reset()` method and that it also affects all the substates.
I tested and it seems to work as desired.  Just in case, I asked about it on the reflex-local-auth repo (as they are not using it in their example, so I wonder if there are some hidden issues with this approach): https://github.com/masenf/reflex-local-auth/issues/26

luator:
Okay, I _did_ just notice an issue with this: Since it clears everything, it also affects state variables which we might want to preserve upon logout.  Specifically, I noticed that when using `reset()`, the language setting is always reset to English when logging out, even if it was German before.  But maybe this could be solved by more carefully setting up the state classes (i.e. session-independent data could be in a separate state class that does not inherit from `SessionState`).

SebastianBoehler:
Good catch, I agree that resetting `SessionState` directly is too broad if it contains state we intentionally want to preserve across logout, like the selected language.

The relevant Reflex docs are here: https://reflex.dev/docs/utility-methods/other-methods/. They describe `reset` as setting all Vars for the given state back to defaults, including substates. So the observed language reset is expected behavior, not a Reflex bug.

For this PR, I think the most pragmatic update would be to preserve the current language explicitly around the reset:

```python
language = self.language
self.do_logout()
self.reset()
self.language = language
```

That keeps the PR surgical and still removes the fragile manual list of page states. I would avoid implementing a custom "reset everything except LanguageState" traversal, because that starts moving back toward custom/manual reset logic.

Longer term, if we find more session-independent state that should survive logout, then I think that is the point where it becomes worth splitting the state tree more carefully: keep preferences/session-independent state outside the subtree we reset on logout, and reset only the user-data/page-cache subtree.

luator:
Sorry, I lost track of that one.   After my review, I asked at the reflex-local-auth repo for some clarification on how they think cleanup during logout should be handled (https://github.com/masenf/reflex-local-auth/issues/26) and wanted to wait for a response there.  Unfortunately, I didn't get one so far.

I'm not so happy with the special handling for the language variable but it might be the best solution for now.  So I would accept the PR if you reopen and apply that change.

PR #365, opened 2026-06-10
Use OPENAI_BASE_URL if specifed

luator:
It can be set as environment variable or in the `.env` file.  If set, its value is used when initializing the OpenAI client.

This can be used for using other LLM providers that support the same API.

I also reduced the output during initialisation a bit.  I don't think we need all the prints just saying that everything is okay.  Instead, only print if something is wrong or noteworthy (e.g. it might be good to mention if a custom base url is set).

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Create local environment file `.env` with the following content:
 
 ```bash
 OPENAI_API_KEY=<your key>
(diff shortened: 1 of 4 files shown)


Andrewtuebingen (approved):
Everything seems to be working as before, thanks!

PR #367, opened 2026-06-11
[codex] migrate env settings to pydantic

### LINKED ISSUE #366: decouple seems to be unmaintained
By coincidence, I noticed that the repo of the 'decouple' package hasn't seen any new commits in the last three years.  There is also a several months old [issue asking about the status](https://github.com/HBNetwork/python-decouple/issues/184) without resonse.

Hence, we might want to consider moving to some other solution for loading config from `.env`.
I found that Pydantic offers something for this: https://pydantic.dev/docs/validation/latest/concepts/pydantic_settings/
It seems quite capable and we already have a dependency on the core Pydantic through Reflex, so adding pydantic-settings hopefully doesn't add to much.

So I would suggest that we move from decouple to pydantic-settings for a more powerful and actively maintained solution.  Since nothing is broken so far, I'd consider this low priority, though.

SebastianBoehler:
Closes #366.

## What changed
- Replaced direct python-decouple usage with a small typed AppSettings helper based on pydantic-settings.
- Kept OPENAI_API_KEY loading from the real environment and the project .env file.
- Removed the unused direct dotenv dependency; pydantic-settings already depends on python-dotenv for dotenv parsing.
- Added focused tests for environment, dotenv, and missing-key behavior.

## Rationale
Pydantic Settings is the better fit here than plain dotenv loading because it provides a typed settings object, validation, environment-variable loading, and .env support in one app-facing API. Plain python-dotenv remains useful for simple env injection, but would leave required-value checks and type conversion scattered through the code.

References checked:
- https://pydantic.dev/docs/validation/latest/concepts/pydantic_settings/
- https://fastapi.tiangolo.com/advanced/settings/
- https://pypi.org/project/python-dotenv/

## Validation
- uv run pytest tests/test_app_settings.py tests/parser_test.py tests/test_example.py
- uv run ruff check
- uv run pyright aitutor/app_settings.py aitutor/aitutor.py aitutor/pages/chat/state.py tests/test_app_settings.py
- uv tree --package pydantic-settings --depth 2 confirmed python-dotenv remains installed transitively.
- uv run reflex run --frontend-port 3100 --backend-port 8100 --backend-host 127.0.0.1 with OPENAI_API_KEY in the process environment started successfully; frontend returned HTTP 200 and backend /ping returned "pong".
- uv run reflex run --frontend-port 3100 --backend-port 8100 --backend-host 127.0.0.1 with a temporary project .env and no exported OPENAI_API_KEY started successfully; startup logged OPENAI_API_KEY found and AppSettings loaded the dummy key from .env.
- Browser check loaded http://localhost:3100/ and rendered the dashboard. The remaining console errors are existing Emotion SSR pseudo-class warnings, not dotenv/settings related.
- GitHub Actions: pyright, pytest, ruff-check, ruff-format, and docker-build passed.


aitutor/app_settings.py
```diff
@@ -0,0 +1,34 @@
...
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+ENV_FILE_PATH = Path(__file__).resolve().parent.parent / ".env"
+
+
+class AppSettings(BaseSettings):
+    """Settings loaded from environment variables and the project `.env` file."""
+
+    model_config = SettingsConfigDict(
+        env_file=ENV_FILE_PATH,
```
luator: The .env file should be searched for in the current working directory, not somewhere relative to this file (the latter will not work in the production setup).  I think just putting `env_file=".env"` should be the way to go here.

luator:
Thanks for the PR.  I'd like to wait for #365 to be merged before doing proper review, as there will be some conflicts.

luator:
#365 is merged now.  It would be great, if you could resolve the conflicts.

SebastianBoehler:
Resolved the merge conflicts

aitutor/aitutor.py
```diff
@@ -164,16 +164,16 @@ def initialize():
...
+    # check if an OpenAI key is configured, if not, we exit
+    settings = get_settings()
+    try:
+        settings.require_openai_api_key()
+    except ValueError as e:
+        print("\033[91m" + str(e) + "\033[0m")
         sys.exit(1)
-    openai_base_url = decouple.config("OPENAI_BASE_URL", cast=str, default="")
+
+    print("OPENAI_API_KEY found in environment variables.")
```
luator: Since the key should always be there, let's remove this print (no error means the key is there, so no need for an explicit message).

aitutor/app_settings.py
```diff
@@ -0,0 +1,40 @@
...
+        env_file_encoding="utf-8",
+        extra="ignore",
+    )
+
+    openai_api_key: str = Field(default="", validation_alias="OPENAI_API_KEY")
+    openai_base_url: str = Field(default="", validation_alias="OPENAI_BASE_URL")
```
luator: - api key is required, so there should be no default here
- default for base_url should be `None`

With those changes, the two getter methods below can be removed and we can just directly access the attributes.

luator (changes requested):

SebastianBoehler:
Thanks for the review. Implemented the comments

aitutor/app_settings.py
```diff
@@ -1,40 +1,38 @@
...
+        default=None, validation_alias="OPENAI_BASE_URL"
+    )
 
-    def optional_openai_base_url(self) -> str | None:
-        """Return a configured OpenAI-compatible base URL, if present."""
-        base_url = self.openai_base_url.strip()
-        return base_url or None
+    @field_validator("openai_base_url", mode="before")
+    @classmethod
+    def empty_openai_base_url_as_none(cls, value: object) -> object:
```
luator: To my understanding, this is automatically called when setting `openai_base_url` and is nothing we expect to be called explicitly in our code?  If so, I'd mark it as private, i.e. prefix the name with and underscore.

luator (commented):
Thanks for the changes.  Just one little request for the field validator method name, the rest looks good.

luator (approved):
Thanks.

PR #369, opened 2026-06-17
Andrew/exercise2

### LINKED ISSUE #319: Req 15: In-lecture role "Tutor" can view submissions of the lecture

### LINKED ISSUE #320: Req 16: In-lecture role "Tutor" can view reports of bad conversations in the lecture

### LINKED ISSUE #359: privacy bug

Andrewtuebingen:
## Summary

This PR adds lecture-specific submission and report handling to the exercise workflow.

## Changes

- Added lecture-specific submissions pages, including filtering and access control for tutors, owners, and admins.
- Added lecture-specific report overview and report detail pages, so reports can be reviewed directly within a lecture context.
- Extended report storage with a `lecture_id`, allowing reports to stay linked to their lecture even if the original exercise is removed.
- Updated chat and finished-view flows to keep track of the lecture context and redirect users back to the correct lecture pages.
- Added lecture-based permission checks for protected pages and state actions.
- Updated the lecture navbar and routes to include submissions and reports.
- Included the required database migration for the new report-to-lecture relation.


aitutor/auth/protection.py
```diff
@@ -7,9 +7,51 @@
...
+    def is_lecture_owner_or_admin(self) -> bool:
+        """Whether the current user may access owner-only pages of this lecture."""
+        return self._has_lecture_role_at_least(LectureRole.OWNER)
+
+    def _has_lecture_role_at_least(self, required_role: LectureRole) -> bool:
+        """Check the current route lecture role against the required lecture role."""
+        if GlobalPermission.ADMIN in self.global_permissions:
+            return True
+        if self.authenticated_user is None or self.authenticated_user.id is None:
+            return False
```
luator, line 37: I'd always put this check at the beginning, i.e. first check if the user is actually logged in before doing anything else.
I assume that it technically doesn't make a difference but it feels more natural and also more fail-safe to me (e.g. assume that `global_permissions` is not cleared on logout for whatever reason).

aitutor/auth/protection.py
```diff
@@ -105,6 +147,72 @@ def protected_page():
...
+                case LectureRole.TUTOR:
+                    lecture_access_cond = (
+                        LectureAccessState.is_lecture_tutor_or_owner_or_admin
+                    )
+                case _:
+                    lecture_access_cond = LectureAccessState.is_lecture_member_or_admin
+
+            if perms_to_check:
+                perm_cond = SessionState.global_permissions.contains(perms_to_check[0])
+                for perm in perms_to_check[1:]:
```
luator: You could initialize `perm_cond` with `False`, then you can simply loop over `perms_to_check` without special handling of the first element.  Since `perms_to_check` defaults to an empty list, you can even omit the `if/else` in that case.  The loop will simply do nothing for an empty list, leaving `perm_cond` at the initial `False`.

aitutor/pages/lecture_report_view/state.py
```diff
@@ -0,0 +1,149 @@
...
+                    role=Role(msg["role"]),
+                    message=msg["content"],
+                    check_passed=msg.get("check_passed", False),
+                )
+                for msg in conversation_data
+            ]
+
+    @rx.event
+    @state_require_role_or_permission(required_role=UserRole.STUDENT)
+    def toggle_looked_at(self):
```
luator: I think we should add an equivalent decorator for in-lecture permissions.  Currently, there is no guard preventing an unauthorized user to execute this function (assuming there is some way they can trigger execution of this function).

Andrewtuebingen:
Great tipps, thanks for the review!

aitutor/pages/lecture_report_view/state.py
Andrewtuebingen: @luator as there are multiple places where I can use the new decorator, should I do it in this pr or rather in the next one? what is better for you as a reviewer?

luator: Probably better to do it in a separate PR (the smaller the individual changes, the easier to review).

luator (approved):
Thanks!

PR #381, opened 2026-07-13
Andrew/lecture prompt

### LINKED ISSUE #358: unique exercise name is global
if there is an exercise called for example math in lecture x then there can't be an exercise with the same name (ex: math) in lecture y.

for me it doesn't seem like a problem because its very simple to choose another name or a slightly longer one, this is why I didn't put time into fixing this "issue". But if you want me to change it @luator , I will do it.

### LINKED ISSUE #380: make token analyzer lecture specific

Andrewtuebingen:
## Summary

This PR adds lecture-specific prompt management and token analyzer, while refining how prompts and exercises are scoped to lectures.

The main changes are:

- Added lecture-specific prompt management.
- Added a lecture-specific token analyzer page.
- Changed exercise title uniqueness from global to lecture-specific.
- Refined prompt name uniqueness so prompts can share names across different scopes/lectures.
- Improved prompt deletion and replacement behavior.
- Updated exercise management UI around prompt selection/imports.
- Extended lecture navigation with additional lecture-specific pages such as settings and token analyzer.

### Lecture-specific prompt management

Lecture-specific prompt management was added so prompts can now be managed in the context of a single lecture.

This includes:

- Loading global prompts and lecture-specific prompts together where needed.
- Allowing lecture-specific prompts to be created, edited, saved, and deleted.
- Supporting a default prompt per lecture.
- Restricting lecture prompt management to users who may manage the lecture's exercises/prompts.

### Prompt name scoping

Prompt name uniqueness was refined.

Previously, prompt names were effectively treated more globally. This PR changes the behavior so that prompt names can be duplicated as long as they belong to different lecture scopes.

The intended behavior is:

- Global prompt names must remain unique among global prompts.
- Lecture-specific prompt names must remain unique within the same lecture.
- Different lectures may use the same prompt name independently.

### Prompt deletion and replacement rules

Prompt deletion behavior was improved to avoid leaving exercises without valid prompts.

When deleting a lecture-specific prompt:

- The prompt must belong to the current lecture.
- A valid replacement prompt must be selected.
- The replacement prompt may be either global or belong to the same lecture.
- Exercises using the deleted prompt are reassigned to the replacement prompt.
- If the deleted prompt was the lecture default prompt, the lecture default is updated as well.

### Exercise title uniqueness per lecture

Exercise titles are no longer globally unique.

Instead, exercise title conflicts are checked within the current lecture. This allows different lectures to contain exercises with the same title, while still preventing duplicate exercise names inside one lecture.

A database migration removes the previous global unique constraint on `exercise.title`.

### Exercise management updates

The exercise management flow was updated to work with lecture-specific prompts and scoped exercise names.

This includes:

- Updated prompt selection for lecture exercises.
- Loading both global prompts and prompts belonging to the current lecture.
- Sorting prompts so defaults appear first.
- Handling prompt name conflicts during exercise import..

### Lecture-specific token analyzer

A new lecture-specific token analyzer page was added.

The page is integrated into the lecture-specific navigation and is only available to users with permission to view lecture submissions/statistics.

### Navigation updates

The lecture-specific navigation was extended.

New or updated navigation entries include:

- Lecture settings instead of edit_lecture.(edit button removed).
- Lecture prompt management.
- Lecture token analyzer.
- Existing lecture management/submission/report tabs continue to be shown based on permissions.

The previous direct edit button flow was replaced by lecture-specific settings navigation.


aitutor/pages/lecture_manage_exercises/state.py
```diff
@@ -148,25 +147,35 @@ def on_load(self):
...
+                        .where(
+                            or_(
+                                Prompt.lecture_id == None,  # noqa: E711
+                                Prompt.lecture_id == lecture_id,
+                            )
+                        )
+                        .order_by(
                             Prompt.is_default_prompt.desc(),  # type: ignore
+                            Prompt.lecture_id.is_not(None).desc(),  # type: ignore
                             Prompt.id,  # type: ignore
```
luator, line 172: What do you think about sorting by name instead of id?  It might make it easier to find a specific prompt if the list is longer.

aitutor/pages/lecture_manage_exercises/state.py
```diff
@@ -148,25 +147,35 @@ def on_load(self):
...
                         )
                     ).all()
                 )
                 or []
             )
+        self._sort_prompts_by_default()
```
luator: The sorting here seems to be redundant with the `order_by` in the query above.  Are both needed?

aitutor/pages/lecture_manage_exercises/state.py
```diff
@@ -230,12 +239,35 @@ def something_is_selected(self) -> bool:
...
+        """Return whether a title conflicts with exercises in this lecture."""
+        if self.current_lecture_id is None:
+            return False
+
+        with rx.session() as session:
+            db_exercises = session.exec(
+                select(Exercise).where(Exercise.lecture_id == self.current_lecture_id)
+            ).all()
+        return any(
+            exercise.id not in exercise_ids and exercise.title in titles
```
luator: Could this filtering be done by the database?  I'm not sure if something as simple as
```python
.where(
    Exercise.lecture_id == self.current_lecture_id,
    Exercise.id not in exercise_ids,
    Exercise.title in titles,
)
```
would maybe work (maybe not it's just an idea).

luator (commented):
I assume there is a lot of code duplication between `lecture_prompts` and the global `prompts` page.  Could they share components and maybe even some state functions?

luator:
Side remark for future PRs:  This PR got quite large.  For faster reviewing, it would be better to split independent changes (e.g. the token analyzer) into separate PRs.  I also increased my working hours to working four days/week since this month, so I can be a bit more responsive now.

aitutor/pages/lecture_manage_exercises/state.py
Andrewtuebingen: You are right that the extra Python sorting was confusing.

The reason i had it is that there are two different “default” concepts in the code:

1. The global prompt manager sets `Prompt.is_default_prompt`. This is a global default prompt.
2. The lecture prompt manager sets `Lecture.default_prompt_id`. This is the prompt that should be preselected when creating a new exercise in a lecture. It can point to either a lecture specific prompt or a global prompt.

The old `_sort_prompts_by_default()` function used the lecture specific default (`Lecture.default_prompt_id`) and moved that prompt to the top of the internal prompt list. After that it prioritized the global default prompt and then sorted by id.

I removed this Python sorting because the lecture default does not need to be moved to the top of the dropdown. It only needs to be preselected when opening the add exercise dialog, and that is handled separately by looking up `Lecture.default_prompt_id` and setting `current_prompt_id`.

The dropdown itself already renders prompts in two visual groups: lecture specific prompts first, then global prompts. With the updated query, prompts are sorted by name instead of id within these groups. If a global prompt is marked as default in the global prompt manager, it appears at the top of the global prompt group, but it does not move above the lecture specific prompts.

Since global exercises are going away, the global “default prompt” may no longer really mean “default for exercise creation”. We could instead treat it as an admin recommended prompt. In that case, the current behavior makes sense: the recommended global prompt is shown first within the global prompt group. As a follow up, I could rename/reframe the global default action to “recommended prompt” and optionally add a small “recommended” badge in the dropdown.

Andrewtuebingen:
> I assume there is a lot of code duplication between `lecture_prompts` and the global `prompts` page. Could they share components and maybe even some state functions?

yeah good idea. I can make a shared component file. Do you want it this PR or in the next?. I wrote it as an issue

aitutor/pages/lecture_manage_exercises/state.py
Andrewtuebingen: great idea

aitutor/pages/lecture_manage_exercises/state.py
luator: > Since global exercises are going away, the global “default prompt” may no longer really mean “default for exercise creation”. We could instead treat it as an admin recommended prompt. In that case, the current behavior makes sense: the recommended global prompt is shown first within the global prompt group. As a follow up, I could rename/reframe the global default action to “recommended prompt” and optionally add a small “recommended” badge in the dropdown.

How about using `Prompt.is_default_prompt` to set `Lecture.default_prompt_id` when creating a new lecture?  Then the lecture starts with a reasonable default and the lecturer can change it later if desired.  This means later changes of the default prompt would not affect existing lectures but I think that's fine.

I think it would be nice to mark the default prompt with a "default" badge in the dropdown.

Both could be done in separate PRs, though.

luator:
> Do you want it this PR or in the next?. I wrote it as an issue

Let's do it in a separate PR to not block other changes of this one from being merged.

luator (approved):
Thanks a lot!

PR #383, opened 2026-07-17
auto-focus login form

luator:
When entering the login page, the username input field is automatically focused (so no need to klick on it).


Andrewtuebingen (approved):
Thanks

PR #348, opened 2026-05-19
Switch to Python 3.14 and bump Reflex to latest release

luator:
I tested a bit and everything seemed to work as before to me but would probably be good if you could also play around a bit before we merge.

Closes #255

PR #384, opened 2026-07-17
Use centralized methods to access route parameters

luator:
Instead of using reflex's magically created state class attributes, access route parameters via some helper functions. This makes the code easier to understand and is also more type-checker friendly (pyright was complaining about all those undefined attributes).

I added two methods:
- `get_route_param_or_default`: For cases where the parameter is optional and a default value is needed.
- `get_route_param_or_error`: For cases where the parameter is required (raises an error if not found).

I tried to keep changes to the other files minimal but did a little bit of restructuring in some cases to make the code simpler.

@Andrewtuebingen This resolves the pyright errors in #348.  Do you think this approach makes sense?

**Do not merge before:**
- #348

### DIFF
--- a/aitutor/auth/protection.py
+++ b/aitutor/auth/protection.py
@@ -39,7 +39,7 @@ class LectureAccessState(SessionState):
             return True
 
         try:
-            lecture_id = int(self.lecture_id)
+            lecture_id = int(self.get_route_param_or_default("lecture_id", default=""))
         except ValueError:
             return False
 
@@ -223,7 +223,9 @@ def state_require_lecture_role(required_role: LectureRole):
                 return
 
             try:
-                lecture_id = int(self.lecture_id)
+                lecture_id = int(
+                    self.get_route_param_or_default("lecture_id", default="")
+                )
             except ValueError:
                 return
 
--- a/aitutor/auth/state.py
+++ b/aitutor/auth/state.py
@@ -158,6 +158,12 @@ class SessionState(reflex_local_auth.LocalAuthState):
             or GlobalPermission.ADMIN in self.global_permissions
         )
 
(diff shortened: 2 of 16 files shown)


Andrewtuebingen:
Great Idea! its working fine but my terminal is getting spammed with these messages, can we mute these warnings?: Warning: Frontend version 0.8.18 for session I0mBKN6ukb6dA27eAAAB does not match the backend version 0.9.7.
DeprecationWarning: RouterData.page has been deprecated in version 0.8.1. Use RouterData.url instead. It will 
be completely removed in 1.0. (aitutor/auth/state.py:189)
DeprecationWarning: RouterData.page has been deprecated in version 0.8.1. Use RouterData.url instead. It will 
be completely removed in 1.0. (aitutor/auth/state.py:191)
DeprecationWarning: RouterData.page has been deprecated in version 0.8.1. Use RouterData.url instead. It will 
be completely removed in 1.0. (aitutor/auth/state.py:172)
DeprecationWarning: RouterData.page has been deprecated in version 0.8.1. Use RouterData.url instead. It will 
be completely removed in 1.0. (aitutor/auth/state.py:173)

Andrewtuebingen (approved):

luator:
Yeah, that's quite annoying.  They already added a deprecation warning but to my understanding `RouterData.url` doesn't even provide access to the parameters yet, so it's not possible to use it yet.
I'll see if I can somehow selectively catch that warning.

luator:
@Andrewtuebingen For some reason ignoring the warning didn't work but I now wrapped the call in a method, so we at least get the warning only once.  It might anyway be better to not suppress it completely, so we don't forget about it being something to keep in mind.
See https://github.com/reflex-dev/reflex/issues/5689 for the related issue on the Reflex repo.

PR #391, opened 2026-07-22
make manageTags lecture specific

### LINKED ISSUE #386: make manage tags lecture specific

Andrewtuebingen:
## Summary

This PR makes exercise tags lecture-specific.

## Changes

- Added a `lecture_id` relation to tags so tags can belong to a specific lecture.
- Updated tag creation, editing, and deletion in lecture exercise management to only affect tags from the current lecture.
- Added a database migration for the new `tag.lecture_id` column and related indexes/foreign key.

## Why

Previously, tags were global across all lectures. This change allows each lecture to manage its own set of tags independently.


aitutor/pages/lecture_manage_exercises/state.py
```diff
@@ -711,9 +720,14 @@ def load_exercises(self):
...
+            self.tag_names = []
+            return
+
         with rx.session() as session:
             # load tags
-            query_tags = select(Tag)
+            query_tags = select(Tag).where(Tag.lecture_id == self.current_lecture_id)
             self.tag_list = sorted(
                 session.exec(query_tags).all(), key=lambda tag: tag.name.lower()
             )
```
luator: This is not part of your change but since you're touching this function anyway:
Sorting could be done by the database.  It would simplify the code a bit and might be more efficient.

aitutor/models.py
```diff
@@ -119,12 +122,16 @@ class Tag(SQLModel, table=True):
     """Tag model for storing allowed tags."""
 
     id: Optional[int] = Field(default=None, primary_key=True)
-    name: str = Field(unique=True, nullable=False, index=True)
+    name: str = Field(nullable=False, index=True)
+    lecture_id: Optional[int] = Field(
+        default=None, foreign_key="lecture.id", ondelete="CASCADE", index=True
+    )
```
luator: Instead of having separate indices on `name` and `lecture_id`, I think it would be better to have a composite one, that takes both fields into account.  Because our queries typically have both together in the where clause.  This could then also be made unique to block accidental addition of duplicate tags on the database level.

luator (changes requested):
Thanks, just one request regarding the index on the table, see below.

luator (approved):
Thanks!

Issue #370, opened 2026-06-18
Safari: Tables rows do not automatically increase when content is extended


Andrewtuebingen:
I accidentally came across this bug, which really weird because it was never like this. Is it also the same for you @luator ?
the buttons are cropped. only if you press twice on the dropdown menu, it becomes normal.

<img width="1310" height="686" alt="Image" src="https://github.com/user-attachments/assets/2a893903-af48-42e9-b60d-04c2b9a422f3" />

luator:
I can't reproduce.  Which browser are you using?  I tested with Vivaldi (chromium-based) and Firefox.  In both it worked as expected for me (using the current main branch).

Andrewtuebingen:
ok interesting, I used safari

Andrewtuebingen:
This problem also occurs in All_lecture. possible solution: accordion

luator:
I renamed the issue to better describe the general problem.
Since this is a Safari issue (not occurring in Firefox or Chromium-based browsers), it is difficult to investigate from Linux.  However, I just found that there is the "Gnome Web" browser (also called "Epiphany"), which uses WebKit (like Safari) and thus shares some of its specific issues.  Indeed, was able to reproduce the issue with that browser.

luator:
@hazem-alabiad Since you have more experience with front-end development: Did you maybe happen to come across this issue already in the past?  The problem is that if we dynamically add content in a table row (like the "save/cancle"-buttons in the example above), most browsers automatically increase the row height to show the new content but in Safari this is not always happening.  As a result, the content is there but not visible to the user as it's cut off.

hazem-alabiad:
@luator Sure, it seems an annoying bug. I fixed such bugs before. I will look into it. Can you assign it to me?

hazem-alabiad:
I looked into the issue, and here is the summary:

It is actually something to do with Safari only; thus, I am not sure if we will do a fix only for Safari. This is the first question. As I can confirm, I could not reproduce it on other web browsers.

## The problem
- In Safari, if we insert dynamic content inside a standard HTML table, there is a potential to trigger a layout bug.
- That is mainly because of the `rx.vstack` element inside the table cell which causes the height of the cell to expand dynamically (to show the buttons)

## Solutions
We have many solutions for such issues
A. we display the buttons next to the dropdown menu by using `rx.hstack` instead
[...]

PR #395, opened 2026-07-23
fix: make confirmation buttons next to the dropdown menu

hazem-alabiad:
## Issue
The cheapest solution is to use a horizontal stack instead of a vertical one -> no dynamic height issues on Safari

## Fixes
https://github.com/martius-lab/ai-tutor/issues/370

### DIFF
--- a/aitutor/pages/lecture_members/state.py
+++ b/aitutor/pages/lecture_members/state.py
@@ -45,6 +45,8 @@ class LectureMembersState(SessionState):
     available_users: list[AvailableLectureUser] = []
     available_user_filter_query: str = ""
     add_member_dialog_is_open: bool = False
+    editing_member_user_id: int | None = None
+    edit_role_dialog_is_open: bool = False
     current_user_lecture_role: int | None = None
 
     @rx.event
@@ -110,6 +112,21 @@ class LectureMembersState(SessionState):
         """Return loaded members as a list for rendering."""
         return list(self.members.values())
 
+    @rx.var
+    def editing_member(self) -> LectureMemberRow | None:
+        """Return the member currently being edited in the edit-role dialog."""
+        if self.editing_member_user_id is None:
+            return None
+        return self.members.get(self.editing_member_user_id)
+
+    @rx.var(initial_value=False)
+    def editing_member_has_changes(self) -> bool:
+        """Return whether the member currently being edited has pending role changes."""
+        member = self.editing_member
+        if member is None:
+            return False
+        return member.selected_role != member.role
+
     @rx.var(initial_value=[])
     def filtered_available_users(self) -> list[AvailableLectureUser]:
         """Return users available to add, filtered by the search query."""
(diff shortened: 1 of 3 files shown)


luator (commented):
I just tested it.  Adding the buttons at the side makes the dropdown move on the horizontal.  I guess this is due to dynamic column widths in the table?  Could we fix this by assigning fixed widths to the second and third column?

hazem-alabiad:
We agreed to implement a pop-up window for the confirmation 
CC @luator

aitutor/pages/lecture_members/state.py
```diff
@@ -149,6 +166,45 @@ def set_add_member_dialog_is_open(self, value: bool):
...
+        """Close the edit role dialog and reset editing state."""
+        self.edit_role_dialog_is_open = False
+        self.editing_member_user_id = None
+
+    @rx.event
+    def set_editing_member_role(self, role_name: str):
+        """Update the selected role for the member currently being edited."""
+        if (
+            self.editing_member_user_id is not None
+            and self.editing_member_user_id in self.members
```
luator, line 176: Is the `is not None` check needed to make pyright happy?  If not, it could potentially be ommitted, as `None in self.members` should also be False.

luator (commented):
Thanks, functionality-wise this looks already great.
I just have some ideas to make the dialog look a bit nicer (feel free to object if you think they are not good):
- Instead of having a extra line for it, the username could maybe be put in the title ("Edit role of {username}").
- The current role could be marked with a "current" badge in the dropdown itself.  See the [prompt selection dropdown](https://github.com/martius-lab/ai-tutor/blob/7ec79150b30f9e60966080b974a92a2393ce8dc2/aitutor/pages/lecture_manage_exercises/components.py#L622) in the exercise editor for an example on how that can be done.
- The text of `LanguageState.roles_description` could be shown in the dialog (either directly or as hover text of some questionmark icon or something like that).

Andrewtuebingen (changes requested):
It looks great! I have two comments on this PR:

1. unused code


I think some of the old inline role change code is now unused after moving the flow into the dialog.

For example, `role_select()` in `components.py` and `set_member_role()` / `cancel_member_role_change()` in `state.py` don't seem to be used anymore. Maybe we can remove them to avoid keeping two versions of the same logic around.


2. Cancel button does not reset the selected role


One small issue I noticed: when I change the role in the dialog and then cancel it, the temporary `selected_role` value is not reset.

Example:
1. Open edit role dialog for a member
2. Change role from `STUDENT` to `TUTOR`
3. Click cancel the dialog
4. Open the dialog again

The select can still show `TUTOR`, even though it was never saved. I think cancel/close should reset `selected_role` back to the current persisted `role`.

aitutor/pages/lecture_members/state.py
hazem-alabiad: Yes, otherwise it shows `bad-index` warning

Andrewtuebingen (approved):

PR #394, opened 2026-07-23
Send emails via SMTP

luator:
I picked up #351 so we can eventually merge it (most of the functionality was already done there).

Changes I made on top of #351 are:
- Refactoring to simplify the configuration of SMTP settings.
- Making SMTP settings optional.  If not set, the email text will be printed to terminal.  This is mostly to not disrupt current development.  The better approach is to run mailpit (see `compose.mailpit.yaml`) to verify emails are properly sent.
- Added basic email validation during the registration
- A few minor changes

PR #396, opened 2026-07-28
Rename AppSettings to EnvSettings

luator:
This is an attempt to reduce the confusion between "config" and "app settings".  The app settings are those values that are read from the environment (either env variables or .env file), so renaming it to "env settings" hopefully makes it clearer.

Further changes:
- Changed all fields to upper case to be consistent with `SmtpSettings`.
- Removed the validation_aliases, which exactly matched the field names and were thus not needed.

**Do not merge before**
- #394

### DIFF
--- a/aitutor/account_emails.py
+++ b/aitutor/account_emails.py
@@ -1,13 +1,13 @@
 """Account welcome email helpers."""
 
-from aitutor.app_settings import get_settings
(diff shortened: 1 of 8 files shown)


hazem-alabiad (approved):
I am fresh here but, all looks good LGTM 🚀

PR #400, opened 2026-07-29
Use state.reset() to clear all data on logout

### LINKED ISSUE #362: Make sure all state classes are listed in SessionState.perform_logout

### LINKED ISSUE #363: More secure data clearing on logout

luator:
Picking up #364.

Instead of manually listing all substates to call their `on_logout`
function, simply call `reset()` on the `SessionState`.  This resets all
base variables to their default, including all substates.

This means that all states that inherit from `SessionState` will be
complete reset upon logout without having to take care of it explicitly.

This means the `on_logout` methods are now obsolete and are removed.

There is currently one value, which we want to preserve upon logout: the
language setting.  Since it's just one flag, I manually restore it after
the reset().  A more proper solution (if we get more such fields in the
future) would be to move them to some separate state class that does not
inherit from `SessionState` (and thus doesn't get reset on logout).

An alternative approach would have been to dynamically get the list of
substates using `self.substates.values()` (see the implementation of the
`reset()` method) and call their `on_logout()` method.  This would allow
for more flexibility (e.g. for cases like language), but still bears the
risk of forgetting to add fields in the `on_logout()` method.  Hence I
decided to go for the fully-automatic `reset()`.

Closes #363 
Closes #362

### DIFF
--- a/aitutor/auth/state.py
+++ b/aitutor/auth/state.py
@@ -74,45 +74,15 @@ class SessionState(reflex_local_auth.LocalAuthState):
         """
         Handles the logout process for the authenticated user.
         """
-        from aitutor import pages
-
-        states = [
-            pages.AllLecturesState,
-            pages.ChatState,
-            pages.EditLectureState,
-            pages.HomeState,
-            pages.ExercisesState,
-            pages.FinishedViewState,
-            pages.FinishedViewTutorState,
-            pages.LectureExercisesState,
-            pages.LectureManageExercisesState,
-            pages.LectureManagePromptsState,
-            pages.LectureManageTagsState,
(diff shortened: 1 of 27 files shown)


Andrewtuebingen (approved):
That's great! Thanks

PR #406, opened 2026-07-30
Review 1

hazem-alabiad:
## Changes 

1. Make the title of the dialog render ("Edit role of {username}") instead of only "Edit role"
2. Display role description
3. Make the current role label in the dropdown instead of the dialog main body.
4. the `Save` button is disabled if no changes made

## Link 
https://github.com/martius-lab/ai-tutor/pull/395#pullrequestreview-4816854168

### Note
I am not sure if u already have a consistent workflow to follow for fonts and colors? I am asking especially for teh code 
```py
font_size="0.9em",
color=rx.color("gray", 11),
```

I also thought about a callout block but was not sure. WHAT DO YOU THINK? @luator

### DIFF
--- a/aitutor/language_state.py
+++ b/aitutor/language_state.py
@@ -1676,6 +1676,23 @@ class BackendTranslations:
         )
 
     @staticmethod
+    def edit_role_title(language: Language, username: str) -> str:
+        return translate(
+            language,
+            de=f"Rolle von {username} bearbeiten",
+            en=f"Edit Role of {username}",
+        )
+
+    @staticmethod
+    def current_tag(language: Language) -> str:
+        return translate(
+            language,
+            de="aktuell",
+            en="current",
+        )
+
+
+    @staticmethod
     def invalid_password(language: Language) -> str:
         return translate(
             language,
--- a/aitutor/language_state.py
+++ b/aitutor/language_state.py
@@ -1502,6 +1502,20 @@ Please test the chat for functionality after changing the model.
             en="This person will be removed from the lecture.",
         )
 
+    @rx.var
+    def label_new_role(self) -> str:
+        return self.translate(
+            de="Neue Rolle:",
+            en="New role:",
+        )
+
+    @rx.var
+    def current_tag(self) -> str:
+        return self.translate(
+            de="aktuell",
+            en="current",
+        )
+
     # Lecture_overview Strings --------------------------------------------------------------------------------
     @rx.var
     def overview(self) -> str:
@@ -1509,35 +1523,6 @@ Please test the chat for functionality after changing the model.
         return self.translate(
             de="Übersicht",
             en="Overview",
-        )
-
-    @rx.var
-    def edit_role(self) -> str:
-        """Edit role string."""
-        return self.translate(
-            de="Rolle Bearbeiten",
-            en="Edit Role",
-        )
-
-    @rx.var
-    def label_username(self) -> str:
-        return self.translate(
-            de="Benutzername:",
-            en="Username:",
-        )
-
-    @rx.var
-    def label_role(self) -> str:
-        return self.translate(
-            de="Aktuelle Rolle:",
-            en="Current Role:",
-        )
-
-    @rx.var
-    def label_new_role(self) -> str:
-        return self.translate(
-            de="Neue Rolle:",
-            en="New role:",
         )
 
 
@@ -1682,15 +1667,6 @@ class BackendTranslations:
             de=f"Rolle von {username} bearbeiten",
             en=f"Edit Role of {username}",
         )
-
-    @staticmethod
-    def current_tag(language: Language) -> str:
-        return translate(
-            language,
-            de="aktuell",
-            en="current",
-        )
-
 
     @staticmethod
     def invalid_password(language: Language) -> str:
--- a/aitutor/language_state.py
+++ b/aitutor/language_state.py
@@ -1435,6 +1435,25 @@ Please test the chat for functionality after changing the model.
         """Localized student role label."""
         return self.translate(de="Student", en="Student")
 
+    @rx.var
+    def lecture_roles_description(self) -> str:
+        return self.translate(
+            de=textwrap.dedent(
+                """
+                - STUDENT: Kann Übungen ansehen und bearbeiten.
+                - TUTOR: Kann zusätzlich Abgaben der Übungen einsehen.
+                - OWNER: Kann alles (Vorlesungseinstellungen ändern, Übungen anlegen, Mitglieder verwalten...).
+            """
+            ),
+            en=textwrap.dedent(
+                """
+                - STUDENT: Can work on exercises.
+                - TUTOR: Can also view exercise submissions.
+                - OWNER: Can do everything (change lecture settings, create exercises, manage members...).
+            """
+            ),
+        )
+
     # All Lectures Strings ------------------------------------------------------------------------
     @rx.var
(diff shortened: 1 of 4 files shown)


aitutor/pages/lecture_members/state.py
```diff
@@ -131,6 +132,38 @@ def editing_member_has_changes(self) -> bool:
...
 
+    @rx.var(initial_value="")
+    def edit_role_title(self) -> str:
+        """Return localized edit role title with username."""
+        username = self.editing_member.username if self.editing_member else ""
+        return translate(
+            self.language,
+            de=f"Rolle von {username} Bearbeiten",
+            en=f"Edit Role of {username}",
+        )
```
luator: Please don't add translations here.  There is a class `BackendTranslations` for that in `language_state.py`.

aitutor/pages/lecture_members/components.py
```diff
@@ -277,34 +277,16 @@ def edit_role_dialog() -> rx.Component:
...
                             rx.text(LS.label_new_role),
                             rx.select(
-                                (
-                                    LectureRole.STUDENT.name,
-                                    LectureRole.TUTOR.name,
-                                    LectureRole.OWNER.name,
-                                ),
-                                value=LectureMembersState.editing_member.selected_role,
+                                items=LectureMembersState.role_option_labels,
+                                value=LectureMembersState.selected_role_option,
```
luator: Instead of adding the "current" in the backend (and then having to strip it off there again), you can use the "low level" API of `rx.select` (see its [documentation](https://reflex.dev/docs/library/forms/select/low), which is a bit hidden).  This allows you to have separate labels and values.  See also the [prompt selection](https://github.com/martius-lab/ai-tutor/blob/7ec79150b30f9e60966080b974a92a2393ce8dc2/aitutor/pages/lecture_manage_exercises/components.py#L637-L647) where we already use this to add badges to entries.

luator (changes requested):

aitutor/pages/lecture_members/components.py
hazem-alabiad: 1. I moved the logic from BE to FE
2. I used `badge` for the `current` label 
3. I refactored the lecture roles to be used from the class method (single source of truth) using `LectureRole.names()` 
4. Removed redundant language computed var

aitutor/language_state.py
```diff
@@ -1490,42 +1490,27 @@ def kick_member_description(self) -> str:
...
     @rx.var
-    def label_new_role(self) -> str:
+    def overview(self) -> str:
+        """Overview label."""
         return self.translate(
-            de="Neue Rolle:",
-            en="New role:",
+            de="Übersicht",
+            en="Overview",
         )
```
hazem-alabiad, line 1541: The removed redundant computed vars

aitutor/language_state.py
```diff
@@ -1490,42 +1490,27 @@ def kick_member_description(self) -> str:
...
         return self.translate(
-            de="Benutzername:",
-            en="Username:",
+            de="Neue Rolle:",
+            en="New role:",
         )
 
     @rx.var
-    def label_role(self) -> str:
+    def current_tag(self) -> str:
```
luator: Maybe better rename this to `current_badge` (or maybe just `current`?) to avoid confusion with the concept of "tags" that we have for exercises.

luator (commented):
What is the commit mentioned in the `.agents/skills/reflex-dev` file about?  (adding that question here as GitHub doesn't let me comment in the file directly for some reason)

aitutor/pages/lecture_members/components.py
```diff
@@ -315,6 +315,13 @@ def edit_role_dialog() -> rx.Component:
                         align="start",
                         width="100%",
                     ),
+                    rx.vstack(
+                        rx.markdown(
+                            LS.roles_description,
```
luator: As discussed in the meeting: This string (`LS.roles_description`) refers to the global user roles (to be removed soon).  We should add a separate string here for the lecture roles (mostly the same but its OWNER instead of ADMIN).

hazem-alabiad:
> What is the commit mentioned in the `.agents/skills/reflex-dev` file about? (adding that question here as GitHub doesn't let me comment in the file directly for some reason)

It has smth to do with the local agents (better ways of writing reflex and docs that this pulled by the agents directly) so, does not affet anything else that your AI coding agent. I can keep it local though.

aitutor/language_state.py
hazem-alabiad: I agree `current` should be good

aitutor/pages/lecture_members/components.py
hazem-alabiad: _
image
_

luator:
> > What is the commit mentioned in the `.agents/skills/reflex-dev` file about? (adding that question here as GitHub doesn't let me comment in the file directly for some reason)
> 
> It has smth to do with the local agents (better ways of writing reflex and docs that this pulled by the agents directly) so, does not affet anything else that your AI coding agent. I can keep it local though.

I was just wondering what commit is this referring to (and from which repo).  Is it something that needs to be updated in the future?

hazem-alabiad:
> > > What is the commit mentioned in the `.agents/skills/reflex-dev` file about? (adding that question here as GitHub doesn't let me comment in the file directly for some reason)
> > 
> > 
> > It has smth to do with the local agents (better ways of writing reflex and docs that this pulled by the agents directly) so, does not affet anything else that your AI coding agent. I can keep it local though.
> 
> I was just wondering what commit is this referring to (and from which repo). Is it something that needs to be updated in the future?

I do not know. It is offered by reflex team. U prefer to keep it local for now? if so we can add it to gitignore?

luator:
> > > > What is the commit mentioned in the `.agents/skills/reflex-dev` file about? (adding that question here as GitHub doesn't let me comment in the file directly for some reason)
> > > 
> > > 
> > > It has smth to do with the local agents (better ways of writing reflex and docs that this pulled by the agents directly) so, does not affet anything else that your AI coding agent. I can keep it local though.
> > 
> > 
> > I was just wondering what commit is this referring to (and from which repo). Is it something that needs to be updated in the future?
> 
> I do not know. It is offered by reflex team. U prefer to keep it local for now? if so we can add it to gitignore?

Ah, I see.  I wasn't aware that they offer this.  I think it's fine to add but would prefer to do it in a separate commit (maybe even separate PR?) with a link to the corresponding Reflex documentation in the commit message.  Then we have a reference where it came from.

hazem-alabiad:
> > > > > What is the commit mentioned in the `.agents/skills/reflex-dev` file about? (adding that question here as GitHub doesn't let me comment in the file directly for some reason)
> > > > 
> > > > 
> > > > It has smth to do with the local agents (better ways of writing reflex and docs that this pulled by the agents directly) so, does not affet anything else that your AI coding agent. I can keep it local though.
> > > 
> > > 
> > > I was just wondering what commit is this referring to (and from which repo). Is it something that needs to be updated in the future?
> > 
> > 
> > I do not know. It is offered by reflex team. U prefer to keep it local for now? if so we can add it to gitignore?
> 
> Ah, I see. I wasn't aware that they offer this. I think it's fine to add but would prefer to do it in a separate commit (maybe even separate PR?) with a link to the corresponding Reflex documentation in the commit message. Then we have a reference where it came from.

As far as I read, `agents` should be ignored as it differs from one user to another. So, I deleted the folder in this PR.

luator (approved):

PR #408, opened 2026-07-31
fix: use proper var operation in lecture cards

### LINKED ISSUE #388: My Lectures: Only show "leave" button for lectures where the user is actually member
This only affects admin users.  They can also enter lectures they are not member of and thus all lectures are listed on the "My Lectures" page.

Currently the "Leave lecture" button is shown for all lectures, also those where the user is not member (and thus cannot leave).  I find this confusing and would rather hide the button for those lectures.

hazem-alabiad:
## Problem
When checking the nullability of objects using `is`, it is gonna compile down to `True` always

## Solution
Use the right operation from the official doc https://reflex.dev/docs/vars/var-operations/#comparisons-and-mathematical-operators

## Changes
image

## Test
1. log in
2. go to lectures page
3. u should see the `leave lecture` buttons only next to joined lectures 

## Fixes
https://github.com/martius-lab/ai-tutor/issues/388

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -152,3 +152,4 @@ tmp/
 *.swp
 
 .claude/settings.local.json
+.agents/
--- a/aitutor/pages/my_lectures/components.py
+++ b/aitutor/pages/my_lectures/components.py
@@ -204,7 +204,7 @@ def lecture_action_buttons(
     return rx.hstack(
         enter_lecture_button(lecture, width=button_width),
         rx.cond(
-            role is not None,
+            role != None,
             leave_lecture_button(lecture, width=button_width),
         ),
         spacing="2",


luator (approved):
Ah, perfect!  I wasn't aware that the cond was already there and just not working correctly.

luator:
Tip: If you write "Fixes #388" in the PR description, the issue will be automatically closed when this PR is merged (works also with "closes" or "resolves").

hazem-alabiad:
> Tip: If you write "Fixes #388" in the PR description, the issue will be automatically closed when this PR is merged (works also with "closes" or "resolves").

I guess we can configure that in teh settings? 

Shall I stop mentioning the task in the PR?

luator:
> > Tip: If you write "Fixes #388" in the PR description, the issue will be automatically closed when this PR is merged (works also with "closes" or "resolves").
> 
> I guess we can configure that in teh settings?
> 
> Shall I stop mentioning the task in the PR?

No, mentioning it is good.  I just commented on it because the format in which you mentioned it, did not trigger the auto-close (it has to be "fixes #xxx" in one row, having it in a "Fixes" section doesn't work).
(In this PR, I added it in the "Development" section, which also triggers the auto-close.)

PR #411, opened 2026-08-05
Use RadixThemesPlugin for setting the theme

luator:
The `theme` argument of `rx.App` is deprecated, one should use the `RadixThemesPlugin` instead.

This changes resolves the deprecation warning.  It should have no effect on the look of the application.


hazem-alabiad (approved):
Great work! LGTM 🚀

PR #412, opened 2026-08-05
Fix db migration issue and set constraint naming conventions

luator:
There was an issue with one of the alembic scripts. It failed on postgres, as the assumed name of the
"unique" constraint ("uq_prompt_name") did not match with what postgres did by default.
Fixed for this instances by getting the actual constrain name from the database (code for this by claude).

To avoid this in the future, I added explicit naming conventions for all our models (see https://alembic.sqlalchemy.org/en/latest/naming.html).  This will only affect newly added constraints/indices, though, so for already existing ones we may still need to add workarounds like above.


Andrewtuebingen (approved):
Thanks!

PR #416, opened 2026-08-07
refactor: delete unsed functions & reset state on cancel

hazem-alabiad:
1. delete unsued code in state file
2. reset selected role on cancellation


Andrewtuebingen (changes requested):
Thanks! The cancel reset issue is fixed, and `role_select()` and `set_member_role()` were removed. However, `cancel_member_role_change()` is still present and appears unused, could we remove that as well?

Andrewtuebingen (approved):
Thanks

luator (approved):

PR #417, opened 2026-08-07
feat: add global announcement banner component and state management

### LINKED ISSUE #344: Add banner for maintenance notifications
Add a way for admins to display banner with some text at the top of all pages.  This could then be used to notify users about upcoming maintenance and the like.

For now, let's keep it simple: A single text field in the global settings and a drop-down menu to choose a colour for the banner.  If the text in the settings is not empty, show it in a box of the specified colour at the top of each page.

hazem-alabiad:
# Summary 
Implement a global banner for announcements like maintenance 

# Changes
1. Implement banner UI and mount it globally on top, below the header
2. Implement banner state file 
3. Implement admin UI page under configuration 
4. connect to DB under `config` table
5. add pre config to config toml file

# Scope 
1. No cancellation button
2. Admin manually enables/disables

# Notes
1. this PR includes DB migration
2. the message body supports `md`

# Testing
1. go to admin settings
2. go to configuration
3. play with the banner section on/off, type, and message
4. it should reflect immieadtly on DB and UI

Screenshot 2026-08-12 at 22 00 37


aitutor/pages/configuration/components.py
```diff
@@ -88,6 +88,65 @@ def info_icon(info_text: str | rx.Var[str]) -> rx.Component:
     )
 
 
+def banner_form_group() -> rx.Component:
+    """Returns a form group card wrapping banner-related controls."""
+    return rx.card(
+        rx.vstack(
+            rx.hstack(
+                rx.heading(LS.banner_section_title, size="5", weight="bold"),
```
luator, line 96: I think by default `rx.heading` produces a `<h1>` heading, so instead of adjusting it's size with `size=`, it's semantically better to set the "heading level" with `as_="h3"` (or whatever is appropriate).

aitutor/pages/configuration/components.py
```diff
@@ -88,6 +88,65 @@ def info_icon(info_text: str | rx.Var[str]) -> rx.Component:
...
+                info_icon(LS.banner_section_info),
+                align="center",
+                spacing="2",
+            ),
+            rx.hstack(
+                rx.text(LS.banner_is_open_label, weight="medium"),
+                rx.switch(
```
luator, line 103: Some time ago we looked in to conventions when to use switch vs checkbox.  The result back than was that a switch should take effect immediately, while for options that only take effect upon explicitly saving (which is the case here), checkbox should be used.
So following that convention, I'd replace the `rx.switch` with `rx.checkbox` here.

aitutor/states/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+    BannerMessageType,
+    BannerState,
+)
+from aitutor.states.config_state import DisplayConfigState
+
+__all__ = [
+    "BannerMessageType",
+    "BannerState",
+    "DisplayConfigState",
+]
```
luator, line 13: Nothing critical, just a suggestion: Instead of the `__all__`, you can do all the imports like this:
```python
from aitutor.states.config_state import DisplayConfigState as DisplayConfigState
```
This also fixes the "unused import" warning but is (in my opinion at least) a bit easier to maintain than the additional `__all__` list.

aitutor/states/banner_state.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @rx.event
+    def refresh_banner(self):
+        """Refresh the banner vars by toggling the trigger_var."""
+        self.trigger_var = not self.trigger_var
+
+    @rx.var
+    def color_scheme(self) -> BannerColor:
+        """Return the corresponding color scheme for the message type."""
+        return BANNER_TYPE_COLORS.get(self.message_type, "blue")
```
luator, line 68: Is there any valid situation in which `self.message_type` could be a value which is not in `BANNER_TYPE_COLORS`?  If not, I'd rather not use `get` so we see the error if something is wrong.

```suggestion
        return BANNER_TYPE_COLORS[self.message_type]
```

aitutor/states/banner_state.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @rx.var
+    def icon(self) -> BannerIcon:
+        """Return the corresponding icon name for the message type."""
+        return BANNER_TYPE_ICONS.get(self.message_type, "info")
```
luator, line 73: same as above

aitutor/config.py
```diff
@@ -99,6 +102,15 @@ def add_configprompts_to_db():
 # Config DB ----------------------------------------------------------------------------
 
 
+def _parse_banner_message_type(val: str | BannerMessageType) -> BannerMessageType:
+    if isinstance(val, BannerMessageType):
+        return val
+    try:
+        return BannerMessageType(val.lower())
+    except ValueError:
+        return BannerMessageType.INFO
```
luator, line 111: Similar question as above: Unless invalid inputs are expected to happen in production, I'd rather let the error escalate instead of silently catching it here.

aitutor/models.py
```diff
@@ -302,6 +313,19 @@ class Config(SQLModel, table=True):
...
     registration_code: str
     exercise_token_limit: int
+    banner_message: str = Field(default="")
+    banner_message_type: BannerMessageType = Field(
+        default=BannerMessageType.INFO,
+        sa_column=Column(
+            sa.Enum(
+                BannerMessageType,
+                values_callable=lambda x: [e.value for e in x],
+            ),
```
luator, line 336: Does specifying a `sa.Enum` like make it behave differently than if we would simply set the type to `BannerMessageType` (we do it like that for other enum fields, see for example `LinkUserLecture.role`).

alembic/versions/88396a0d3bb1_add_banner_to_config.py
```diff
@@ -0,0 +1,40 @@
...
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('config', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('banner_message', sqlmodel.sql.sqltypes.AutoString(), server_default=sa.text("('')"), nullable=False))
+        batch_op.add_column(sa.Column('banner_message_type', sa.Enum('INFO', 'WARNING', 'ERROR', 'SUCCESS', name='bannermessagetype'), server_default=sa.text("'info'"), nullable=False))
+        batch_op.add_column(sa.Column('banner_is_open', sa.Boolean(), server_default=sa.text('0'), nullable=False))
```
luator, line 27:
```suggestion
        batch_op.add_column(sa.Column('banner_is_open', sa.Boolean(), server_default=sa.sql.false(), nullable=False))
```
Otherwise it will not work for PostgreSQL (no idea why Alembic isn't using that by default).

default_config.toml
```diff
@@ -13,6 +13,13 @@ response_ai_model = "gpt-4.1-mini"
 check_ai_model = "gpt-4.1"
 exercise_token_limit = 100_000
 
+# --------------------------
+# banner 
+# --------------------------
+banner_message  = "System maintenance is scheduled for tonight at 10 PM UTC."
+banner_message_type = "WARNING"
+banner_is_open = true
```
luator, line 21: Since these values will be used when initializing a new instance, I'd rather disable it by default.

luator (changes requested):

aitutor/pages/configuration/components.py
hazem-alabiad: True, does nto make sense to be `h1` 
I made it as `h2`

aitutor/pages/configuration/components.py
hazem-alabiad: Makes lots of sense.

aitutor/states/banner_state.py
hazem-alabiad: I agree. When we fill the value of a type we even guard the input value so the user cannot enter any value except the ones that we specify, so we can guarantee no indexing errors.

aitutor/models.py
hazem-alabiad: simplified it

Andrewtuebingen (changes requested):
It looks great! please add the SQLite foreign-key handling required by `docs/contribute.md` for batch-altering existing tables.

hazem-alabiad:
> It looks great! please add the SQLite foreign-key handling required by `docs/contribute.md` for batch-altering existing tables.

Already added

Andrewtuebingen (changes requested):
Thanks, on Sqlite it's working great, but on Postgresql it's not running. Is it working for you on postgresql @hazem-alabiad ? The problem is `88396a0d3bb1_add_banner_to_config.py` adds a column using the native enum type `bannermessagetype`, but the enum type is never created first. Running `uv run reflex db migrate` raises `psycopg.errors.UndefinedObject: type "bannermessagetype" does not exist`.

Andrewtuebingen:
I have tried for a while to find a solution and my new commit seems to be working on Postgres. But @luator knows Postgres better than me, so do you think this makes sense?

Andrewtuebingen (approved):
looks great! I just fixed a small revision mistake

luator (approved):

PR #434, opened 2026-08-12
chore: Add test to run `reflex db migrate`

luator:
This is to make sure that the alembic files are in a proper state.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,9 @@ on:
       - main
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   pytest:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,3 +21,17 @@ jobs:
 
       - name: Run tests
         run: uv run pytest tests/
+
+  db_migrate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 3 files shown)


hazem-alabiad:
The pipeline is failing. I guess the stacking is flipped? 
I believe its the same issue; however, the migration is not fixed yet here.

luator:
> The pipeline is failing. I guess the stacking is flipped?

That's intentional.  I first added the test, so I could verify that with the current state it fails as it should.  Only after that I looked into fixing it.

.github/workflows/test.yml
```diff
@@ -21,3 +24,17 @@ jobs:
...
       - name: Run tests
         run: uv run pytest tests/
+
+  db_migrate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
hazem-alabiad, line 34: One tiny note to enable teh cache https://docs.astral.sh/uv/guides/integration/github/#caching

hazem-alabiad, line 34: `enable-cache: true`

hazem-alabiad (approved):
Other than that tiny note all looks good

.github/workflows/test.yml
luator: According to the README of the setup-uv action, caching is already enabled by default.

PR #435, opened 2026-08-12
fix: alembic merge heads

luator:
We somehow ended up with a branched alembic history (probably due to merging two branches that were developed in parallel).  Due to this it wasn't possible anymore to run `reflex db migrate`.

Fixed by calling
```
alembic merge heads
```


alembic/versions/dc5a36ddd68a_.py
```diff
@@ -0,0 +1,28 @@
+"""empty message
```
hazem-alabiad: Maybe adding a meaningful message would be great? in case of later look up

hazem-alabiad (approved):
LGTM

PR #438, opened 2026-08-12
Add GitHub workflow to check `reflex db makemigrations`

luator:
- Move the reflex database checks into a separate file
- Add a check verifying that `reflex db makemigrations` has no output.

This is to avoid that some change on the models is merged without the corresponding migration script.


hazem-alabiad (approved):
Looks good to me. I am not that expert with DB tho :)

PR #442, opened 2026-08-25
Admin UI to create tokens for magic lecturer registration links

luator:
Add a section to the configuration page where "lecturer registration tokens" can be created (and deleted).
On creation, an expiration date has to be set after which the token will be invalid (but not automatically deleted).

The purpose of these tokens is to provide a "magic" link to the registration page through which the new account directly gets the "lecturer" permission.  This is to make the onboarding process for lecturers easier (without this, they would have to create an account, then notify an admin and wait for the permission to be granted by the admin).

This PR only includes the token management.  The corresponding change on the registration process will be added in a separate (stacked) one, for easier review.


Andrewtuebingen (approved):
Thank you! As I understand it / at the moment, registration still works when the link contains an expired or already deleted token because the `lrt` parameter is not evaluated yet. I assume the follow up PR will validate the token and reject registrations via expired, deleted, or otherwise invalid tokens right?

luator:
> I assume the follow up PR will validate the token and reject registrations via expired, deleted, or otherwise invalid tokens right?

Not quite. I kept it rather simple: An error is shown above the registration form that the token is invalid but registration is still possible (will simply result in a normal account).  Maybe not ideal but let's rather discuss this in the other PR if you think an invalid token should block the registration.

luator:
I resolved merge conflicts and made some minor layout changes, so that the cards have the same width.

PR #443, opened 2026-08-25
Grant 'lecturer' permission to accounts created via magic link

luator:
When a valid lecturer registration token is passed to the 'register' page as `?lrt=...` parameter, the new account will automatically be granted the 'lecturer' permission.

### DIFF
--- a/aitutor/language_state.py
+++ b/aitutor/language_state.py
@@ -1294,10 +1294,24 @@ Please test the chat for functionality after changing the model.
         )
 
     @rx.var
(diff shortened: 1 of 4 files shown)


Andrewtuebingen (approved):
Great work, thanks

luator:
As discussed in the meeting, I changed it to not show the form anymore when the token is invalid.  This means it is not possible to accidentally register a non-lecturer account with a bad token just by ignoring the error.

Screenshot_20260901_150551

Andrewtuebingen (approved):
looks great

PR #445, opened 2026-08-26
fix (alembic): Do not add unique constraint, it already exists

luator:
Probably a conflict caused by commit 64e420cd1d, where the constraint name was explicitly added in an older migration file. Not sure why it wasn't an issue for SQLite but PostgreSQL complained about the already existing constraint.

I also added a check to the GitHub workflows to catch this kind of issue in the future.


hazem-alabiad (approved):
thankss

PR #446, opened 2026-08-26
chore: Test database migration in Docker setup (with postgres)

luator:


luator:
This is added to #445, which also fixes the failing test.

PR #455, opened 2026-08-27
feat: limit the text input from users in exercise management page

### LINKED ISSUE #450: Exercise management

hazem-alabiad:
## Summary
Limit the text input from teh user to prevent harmful actions

## Changes

Added `max_length` properties to the following exercise and tag form fields:

| Field Name | Max Length / Limit |
|:---|:---|
| **Title** (`title`) | `200` characters |
| **Description** (`description`) | `5,000` characters |
| **Lesson Context** (`lesson_context`) | `10,000` characters |
| **Days to Complete** (`days_to_complete`) | `4` digits |
| **Tag Name** (`tag_name` in new & rename dialogs) | `50` characters |

## Testing
- go to `lectures/manage_exercises`
- try to insert text beyond the limits above 
- you should not be able to

### DIFF
--- a/aitutor/pages/lecture_manage_exercises/components.py
+++ b/aitutor/pages/lecture_manage_exercises/components.py
@@ -48,6 +48,7 @@ def new_tag_dialog():
                         ManageTagsState.add_new_tag,
                         None,
                     ),
+                    max_length=50,
                 ),
                 rx.hstack(
                     rx.dialog.close(
@@ -97,6 +98,7 @@ def rename_tag_dialog():
                         ManageTagsState.edit_tag_name(),
                         None,
                     ),
+                    max_length=50,
                 ),
                 rx.hstack(
                     rx.dialog.close(
@@ -775,6 +777,7 @@ def add_edit_exercise_form(mode: DialogMode) -> Sequence[rx.Component]:
             width="100%",
             type="text",
             name="title",
+            max_length=200,
         ),
         # description
         rx.text(
@@ -798,6 +801,7 @@ def add_edit_exercise_form(mode: DialogMode) -> Sequence[rx.Component]:
             height="150px",
             type="text",
             name="description",
+            max_length=5_000,
         ),
         # lesson context
         rx.text(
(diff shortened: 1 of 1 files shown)


aitutor/pages/lecture_manage_exercises/components.py
```diff
@@ -818,6 +822,7 @@ def add_edit_exercise_form(mode: DialogMode) -> Sequence[rx.Component]:
             height="200px",
             type="text",
             name="lesson_context",
+            max_length=10_000,
```
luator: The limit for the lesson context should probably be higher.  I just looked at one randomly picked exercise from the past semester and the context had ~7500 characters, so not too far from 10 000.
Let's use something generous there for now (maybe 100 000 or even more).  The purpose of those limits is mostly to prevent users from choking the server by sending gigabytes of data.

luator (changes requested):
Since those are only frontend limits, I assume they could easily be circumvented by a malicious user.  So we should also add corresponding checks in the backend.

aitutor/pages/lecture_manage_exercises/state.py
```diff
@@ -573,6 +585,16 @@ def add_exercise(self, form_data: dict):
...
         with rx.session() as session:
+            form_data["title"] = form_data["title"][
+                : LECTURE_MANAGE_EXERCISES_FIELD_MAX_LENGTHS["title"]
+            ]
+            form_data["description"] = form_data["description"][
+                : LECTURE_MANAGE_EXERCISES_FIELD_MAX_LENGTHS["description"]
+            ]
+            self.lesson_context = self.lesson_context[
+                : LECTURE_MANAGE_EXERCISES_FIELD_MAX_LENGTHS["lesson_context"]
+            ]
```
luator: lesson_context is already clipped in it's setter function, I guess that's sufficient?

aitutor/pages/lecture_manage_exercises/state.py
```diff
@@ -735,6 +757,16 @@ def toggle_visibility(self, exercise: Exercise):
...
```
luator: same as above

luator (approved):

PR #458, opened 2026-08-27
feat: limit the text input from users in user and member management pages

### LINKED ISSUE #453: user & member managment

hazem-alabiad:
## Summary

Limit the text input from the user to prevent harmful actions, enforce standard email/username boundaries, and ensure clean database records.

Added frontend input limits (`max_length`) and corresponding backend checks/clamping (`*_FIELD_MAX_LENGTHS`) `[NEW]`.
Centralized input field length constants into `aitutor.global_vars` (`GV.*`) to avoid mismatches across pages `[NEW]`.

## Changes

Added length limits to the following user and member management form fields:

| Field Name | Max Length / Limit |
|:---|:---|
| **Username** (`username`) | `100` characters (`GV.USERNAME_MAX_LEN`) |
| **Email** (`email`) | `254` characters (`GV.EMAIL_MAX_LEN`) |
| **Password & Confirm Password** (`password`) | `128` characters (`GV.PASSWORD_MAX_LEN`) |
| **New Password** (`new_password`) `[NEW]` | `128` characters (`GV.PASSWORD_MAX_LEN`) |
| **Registration Code** (`registration_code`) | `150` characters (`GV.REGISTRATION_MAX_LEN`) |
| **Available User Search Filter** (`available_user_filter_query`) | `100` characters (`GV.USERNAME_MAX_LEN`) |

- `[NEW]` Centralized authentication and input limit constants in `aitutor/global_vars.py` (`USERNAME_MAX_LEN`, `EMAIL_MAX_LEN`, `PASSWORD_MAX_LEN`, `REGISTRATION_MAX_LEN`).
- `[NEW]` Updated `aitutor/pages/configuration/state.py` (file in main) to use `GV.REGISTRATION_MAX_LEN` to keep limits synchronized across the app.
- `[NEW]` Added `max_length=128` to **New Password** in the edit user dialog.
- `[NEW]` Backend checks: Added length validation and clamping (`*_FIELD_MAX_LENGTHS`) across all fields.

## Testing

- go to `/register` and test input limits on username, email, password, and registration code
- go to `/admin_settings/manage_users`, click edit on a user, and verify username, email, and new password inputs block text exceeding the limits
- go to `/lectures/members/[lecture_id]`, open Add Member dialog, and test the user search input limit (100 chars)
- go to `/admin_settings/configuration` and verify registration code field still works with the global limit
- you should not be able to exceed these limits

### DIFF
--- a/aitutor/global_vars.py
+++ b/aitutor/global_vars.py
@@ -20,3 +20,12 @@ CHAT_MESSAGE_CHAR_LIMIT = 15000
 
 """colors"""
 GREEN_CHECK_COLOR = rx.color("green", 9)
+
+
+# --- max length for UI input fields ---
+
+# Auth
+USERNAME_MAX_LEN = 100
+EMAIL_MAX_LEN = 254
+PASSWORD_MAX_LEN = 128
+REGISTRATION_MAX_LEN = 150
--- a/aitutor/pages/configuration/state.py
+++ b/aitutor/pages/configuration/state.py
@@ -7,6 +7,7 @@ from zoneinfo import ZoneInfo
 import reflex as rx
 from sqlmodel import select
 
+import aitutor.global_vars as GV
 from aitutor.auth.protection import state_require_role_or_permission
 from aitutor.auth.state import SessionState
 from aitutor.global_vars import TIME_ZONE
@@ -26,7 +27,7 @@ from aitutor.states.banner_state import (
 from aitutor.states.config_state import DisplayConfigState
 
 CONFIG_FIELD_MAX_LENGTHS: dict[str, int] = {
-    "registration_code": 150,
+    "registration_code": GV.REGISTRATION_MAX_LEN,
     "response_ai_model": 100,
     "check_ai_model": 100,
     "how_to_use_text": 10_000,
--- a/aitutor/pages/login_and_registration/components.py
+++ b/aitutor/pages/login_and_registration/components.py
@@ -7,7 +7,10 @@ from reflex_local_auth.pages.components import MIN_WIDTH
 from aitutor import components, routes
 from aitutor.language_state import LanguageState
 from aitutor.pages.legal_infos.loader_functions import get_privacy_notice_short
-from aitutor.pages.login_and_registration.state import MyRegisterState
+from aitutor.pages.login_and_registration.state import (
+    AUTH_FIELD_MAX_LENGTHS,
+    MyRegisterState,
+)
 
 
 def input(name, placeholder, **props) -> rx.Component:
@@ -167,6 +170,7 @@ def register_form() -> rx.Component:
                 value=MyRegisterState.username,
                 on_change=MyRegisterState.set_username,
                 disabled=MyRegisterState.registration_in_progress,
+                max_length=AUTH_FIELD_MAX_LENGTHS["username"],
             ),
             rx.text(LanguageState.email),
             input(
@@ -178,6 +182,7 @@ def register_form() -> rx.Component:
                 value=MyRegisterState.email,
                 on_change=MyRegisterState.set_email,
                 disabled=MyRegisterState.registration_in_progress,
+                max_length=AUTH_FIELD_MAX_LENGTHS["email"],
             ),
             rx.text(LanguageState.password),
             password_input(
@@ -188,6 +193,7 @@ def register_form() -> rx.Component:
                 value=MyRegisterState.password,
                 on_change=MyRegisterState.set_password,
                 disabled=MyRegisterState.registration_in_progress,
+                max_length=AUTH_FIELD_MAX_LENGTHS["password"],
             ),
             rx.text(LanguageState.confirm_password),
             password_input(
@@ -198,6 +204,7 @@ def register_form() -> rx.Component:
                 value=MyRegisterState.confirm_password,
                 on_change=MyRegisterState.set_confirm_password,
                 disabled=MyRegisterState.registration_in_progress,
+                max_length=AUTH_FIELD_MAX_LENGTHS["confirm_password"],
             ),
             rx.cond(
                 MyRegisterState.needs_registration_code,
@@ -210,6 +217,7 @@ def register_form() -> rx.Component:
                         value=MyRegisterState.registration_code,
                         on_change=MyRegisterState.set_registration_code,
                         disabled=MyRegisterState.registration_in_progress,
+                        max_length=AUTH_FIELD_MAX_LENGTHS["registration_code"],
                     ),
                 ),
             ),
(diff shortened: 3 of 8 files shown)


Andrewtuebingen (approved):
Thanks! can you also put a field char limit for logging in (username and password)

Andrewtuebingen (changes requested):
the backend checks are missing

aitutor/pages/login_and_registration/components.py
```diff
@@ -210,6 +217,7 @@ def register_form() -> rx.Component:
                         value=MyRegisterState.registration_code,
                         on_change=MyRegisterState.set_registration_code,
                         disabled=MyRegisterState.registration_in_progress,
+                        max_length=AUTH_FIELD_MAX_LENGTHS["registration_code"],
```
luator: To avoid potential mismatches, I think it would be better to use `CONFIG_FIELD_MAX_LENGTHS["registration_code"]` (from #459) here.

aitutor/pages/manage_users/state.py
```diff
@@ -17,6 +17,12 @@
     UserRole,
 )
 
+MANAGE_USERS_FIELD_MAX_LENGTHS: dict[str, int] = {
+    "username": 100,
+    "email": 254,
+    "new_password": 128,
+}
```
luator: Like with the registration code above, sharing the same variable between here and the registration form would avoid potential mismatches in the future (when changing a value in one place but forgetting to also change the corresponding one of the other page).

I would also increase the limit for the password, maybe double it.  People using password managers may do crazy things :).

The limit for the email seems a bit random.  Why not 256?  It doesn't really matter, I'm just curious.

luator (changes requested):

aitutor/pages/login_and_registration/components.py
hazem-alabiad: That makes sense, lemem figure out global fields and globalize them

aitutor/pages/manage_users/state.py
hazem-alabiad: I thought there must be a standard or convention for this. Turns out there is a standard following the specifications or best practcies 
 https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address
https://github.com/OWASP/ASVS/blob/v4.0.3/4.0/en/0x11-V2-Authentication.md#v21-password-security

but ofc this does not mean we shoudl follow strictly :) we can stil lchange teh numbers

hazem-alabiad: Also, for many password managers, they generate max 128 https://bitwarden.com/blog/how-long-should-my-password-be/

luator: Ah, I wasn't aware that there is an official limit for email addresses.
Regarding passwords: KeepassXC doesn't seem to have a limit but I just tried and 128 still fits a 15-word passphrase using the EFF list.  I guess that's indeed long enough.

aitutor/pages/configuration/state.py
```diff
@@ -7,6 +7,7 @@
 import reflex as rx
 from sqlmodel import select
 
+import aitutor.global_vars as GV
```
luator: Following Python conventions, module names shouldn't use upper case.  So better use `gv` here.

aitutor/global_vars.py
```diff
@@ -20,3 +20,12 @@
...
 GREEN_CHECK_COLOR = rx.color("green", 9)
+
+
+# --- max length for UI input fields ---
+
+# Auth
+USERNAME_MAX_LEN = 100
+EMAIL_MAX_LEN = 254
+PASSWORD_MAX_LEN = 128
+REGISTRATION_MAX_LEN = 150
```
luator:
```suggestion
REGISTRATION_CODE_MAX_LEN = 150
```
With just "REGISTRATION" it is a bit unclear what is meant.

aitutor/gv.py
luator: Please keep the original name of the file, just `gv` is very unclear.  We can use `import ... as gv` to keep the name short in the code.

luator (changes requested):

aitutor/global_vars.py
hazem-alabiad: reverted back to `global_vars.py`

luator (approved):
Thanks!

Andrewtuebingen (changes requested):
Thanks! only one thing: the input limits look good overall, but `PASSWORD_MAX_LEN = 128` is incompatible with the current bcrypt implementation. bcrypt rejects passwords longer than __72 UTF-8 bytes__, so passwords between 73–128 ASCII characters or shorter Unicode passwords, can raise a `ValueError` during registration or an admin password reset. Please validate the encoded byte length and return a user facing error instead of silently truncating the password.

try this as password to see the problem: 😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀

hazem-alabiad:
> Thanks! only one thing: the input limits look good overall, but `PASSWORD_MAX_LEN = 128` is incompatible with the current bcrypt implementation. bcrypt rejects passwords longer than **72 UTF-8 bytes**, so passwords between 73–128 ASCII characters or shorter Unicode passwords, can raise a `ValueError` during registration or an admin password reset. Please validate the encoded byte length and return a user facing error instead of silently truncating the password.
> 
> try this as password to see the problem: 😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀

Good catch! I did not know about this limitation in bcrypt. I will check the byte length of the password and return pronbably a toast error.

hazem-alabiad:
> 😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀

image

hazem-alabiad:
I handled the toast error with a custom message in the registration; however, in the login, we still print the general error message as its a library and to modify the error we need to add way more code. 
I think this issue (input validation) can be revised later in a separate task to enhance the ui and feedback. WHAT DO YOU THINK? @luator @Andrewtuebingen

Andrewtuebingen:
> I handled the toast error with a custom message in the registration; however, in the login, we still print the general error message as its a library and to modify the error we need to add way more code. I think this issue (input validation) can be revised later in a separate task to enhance the ui and feedback. WHAT DO YOU THINK? @luator @Andrewtuebingen

I think it's fine like this, it's not a necessary change in my opinion. Maybe in the future like you said

Andrewtuebingen (approved):
Thanks!

PR #460, opened 2026-08-30
refactor: Replace role-based permission checks with lecture-specific …

### LINKED ISSUE #371: new in-lecture permissions decorator
finish using new equivalent decorator for in-lecture permissions.

### LINKED ISSUE #375: translation missing in manage exercise + auth.py

Andrewtuebingen:
- Replace role-based permission checks with lecture-specific one
- fix translation issues


aitutor/pages/lecture_manage_exercises/state.py
```diff
@@ -132,7 +132,7 @@ def set_exercise_is_selected(self, exercise_id: int | None, is_selected: bool):
         self.exercise_is_selected[exercise_id] = is_selected  # type: ignore
 
     @rx.event
-    @state_require_role_or_permission(required_role=UserRole.STUDENT)
+    @state_require_lecture_role(LectureRole.TUTOR)
```
luator: Tutors should only get access to review submissions, not to edit exercises or any settings of the lecture.

luator (commented):
Looks good implementation-wise.  We should clarify, though, which permissions tutors get.  My plan was to allow tutors only to review submissions (and maybe reports?) but all "edit"-pages (manage exercises, prompts, settings, users) should be limited to owners.
It might make sense to have another role in between TUTOR and OWNER but I'm not sure if that is really needed in practice, so I would only add it, if users actually ask for it.

Andrewtuebingen:
> Looks good implementation-wise. We should clarify, though, which permissions tutors get. My plan was to allow tutors only to review submissions (and maybe reports?) but all "edit"-pages (manage exercises, prompts, settings, users) should be limited to owners. It might make sense to have another role in between TUTOR and OWNER but I'm not sure if that is really needed in practice, so I would only add it, if users actually ask for it.

I thought tutors are the ones that do the whole work and owners are the lecturers that might only create the lecture but then do fewer things in general. because if that's the case, then tutor should be able to do more than just reviewing. Do you know who usually are tutor and who usually are admins/owner in practice

luator:
I guess it differs from lecture to lecture (also depending on the size).  I was thinking of tutors like the ones that correct exercise sheets (typically hiwis) but who don't typically create the exercises.  Creation of exercises, I assume to be done either by the lecturer themselves or a PhD student of the lecturer, who would hopefully be trusted enough, that they could be granted the owner role.

Andrewtuebingen:
I updated the role tutor @luator

luator (approved):
Thanks!

PR #465, opened 2026-09-02
Bump version to 2.0.0a1 and show version in footer

luator:
- Add a helper script `version.sh` to more easily update the version.  It runs `uv version` and updates a file `aitutor/_version.py`, so we can also access the version from within the code.
- Bump version to 2.0.0a1
- Show the version in the footer of all pages, so we can easily see which version is running, e.g. on a server.

### DIFF
--- a/aitutor/_version.py
+++ b/aitutor/_version.py
@@ -1,0 +1,1 @@
+version: str = "2.0.0a1"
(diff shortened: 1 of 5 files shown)


luator:
@Andrewtuebingen @hazem-alabiad I'm already merging, so I can move on with the deployment but let me know if you spot any issue, then I'll fix it in a new PR.

